PR #5, opened 2015-11-18
Update AssemblyInfo.cs to update the version number to 7.0.11

amgdy:
Update the version number to 7.0.11


aidanbebbington:
Thanks for that - it's not been merged, but I've effectively done this myself.  It just took me so long that we've now moved on a few versions :)

PR #12, opened 2016-08-03
Updated metadata to 7.5.1

jzelos:
Nunit passed 233 tests and failed 8 both prior and subsequent to updating the metadata files.

I've just updated the XML files and left everything else as is.


aidanbebbington:
Thanks for sorting that out, must do this more regularly, it's up on nuget.com now.

PR #21, opened 2017-02-26
Update to libphonenumber v8.3.1

omerts:


aidanbebbington:
Thanks for that!  I've updated the nuget package

Issue #13, opened 2016-08-05
NuGet for .NET Standard support


nealculiner:
Hello,

I tried updating a PCL project to .NET standard after removing this package and then tried reinstalling this package from nuget after changing to .NET standard.  This package will not install for some reason.  FYI

https://docs.microsoft.com/en-us/dotnet/articles/standard/library

Thank you for this package by the way!!!

aidanbebbington:
Hi - I'm afraid I've left .NET somewhat, so I'm a bit out of date now.  Do you get a specific error message?

nealculiner:
I created a new class library for .NET core (same as .net standard) and the error is shown below. 

GET https://api.nuget.org/v3-flatcontainer/libphonenumber-csharp/7.5.1/libphonenumber-csharp.7.5.1.nupkg
  OK https://api.nuget.org/v3-flatcontainer/libphonenumber-csharp/7.5.1/libphonenumber-csharp.7.5.1.nupkg 296ms
Installing libphonenumber-csharp 7.5.1.
Installing NuGet package libphonenumber-csharp.7.5.1.
Successfully installed 'libphonenumber-csharp 7.5.1' to ClassLibrary1
========== Finished ==========
PATH=.\node_modules.bin;C:\Program Files (x86)\Microsoft Visual Studio 14.0\Web\External;%PATH%;C:\Program Files (x86)\Microsoft Visual Studio 14.0\Web\External\git
C:\Program Files\dotnet\dotnet.exe restore "D:\Dev\ClassLibrary1.vs\restore.dg"
log  : Restoring packages for D:\Dev\ClassLibrary1\src\ClassLibrary1\project.json...
error: Package libphonenumber-csharp 7.5.1 is not compatible with netstandard1.5 (.NETStandard,Version=v1.5). Package libphonenumber-csharp 7.5.1 supports: net (.NETFramework,Version=v0.0)
error: One or more packages are incompatible with .NETStandard,Version=v1.5.
info : Committing restore...
log  : Writing lock file to disk. Path: D:\Dev\ClassLibrary1\src\ClassLibrary1\project.lock.json
log  : D:\Dev\ClassLibrary1\src\ClassLibrary1\ClassLibrary1.xproj
log  : Restore failed in 374ms.
Errors in D:\Dev\ClassLibrary1\src\ClassLibrary1\ClassLibrary1.xproj
[...]

Issue #6, opened 2015-11-30
.NET Core Support??


cindro:
I recently upgraded an app of mine to ASP.NET 5 .NET Core, then realised that this great library didn't support DNXCore.

Is this support coming soon?

aidanbebbington:
That's no good at all, hadn't caught up to the world of 5.0 to be honest.
What do you expect in terms of support? Does it simply not build or is
there something specific?

cindro:
This is a great library, been using it for while now. I am now upgrading one of my apps but unable to utilise libphonenumber-csharp, I get below error.

The dependency libphonenumber-csharp 7.0.11 does not support framework DNXCore,Version=v5.0

There is no great alternative unfortunately.

FanMix:
.NET Core already here ! It's a great port but still .NET Core is not supported, pity...

aidanbebbington:
Would love to support it.  I'm a bit out of touch with the latest .NET these days, so if someone could do a pull request I would be happy to take it.

cindro:
@FanMix for now you can use https://github.com/jarroda/libphonenumber-dotnet (.NET Core is supported)

jarroda:
I will try to take some time to get my fork updated and issue a PR back to this repo.  I had always intended to but got sidetracked with work.

twcclegg:
If any issues with .net core are found, please let me know.

PR #22, opened 2017-03-11
Port to netstandard 1.0

jarroda:
This PR ports the project to netstandard 1.0.  The library is configured to compile to both .net 4.5 and netstandard 1.0, so backwards compatibility should be maintained. 

Additionally I ported the test project over and all tests pass that did before, with the same 18 tests failing.  

I am using the .NET Core 1.0.0 SDK - Preview 2.1 (build 003177) tooling.

Once merged, this should address #6 and #13


RichiCoder1:
Is there a reason you're not using the latest tools? Project.json is officially deprecated as of this point.

jarroda:
@RichiCoder1 Yeah the reason is that I do all my work on macos using VSCode and CLI tooling, and my attempts at using the msbuild based tooling (a couple of weeks ago) found it to be bug-filled and woefully unproductive.  I realize a new release was put out on the 7th, but I haven't had time to mess with it.  

If you'd like to convert it to the most recent tooling, feel free to pitch in.  Opening the solution in VS 2017 will convert it automatically I presume.

RichiCoder1:
Yup. There's also `dotnet migrate`. Luckily the auto conversion has much improved since then. Will look into it.

aidanbebbington:
Thanks for this!  I've just opened it up in VS2017 which converted things it seems.  I've done some basic smoke tests and updated the nuget package.  Let me know if it doesn't work for you.

PR #29, opened 2017-07-06
Update C# port for number/metadata schema changes <=8.6.0

twcclegg:


aidanbebbington:
damn, you seem pretty keen!  I've granted you push access to this repo if you like.  Or we can change yours to the upstream repo, whatever makes sense.  To be honest I didn't actually write this thing in the first place, I just moved it to Git and updated the metadata a few times.  Is your nuget username the same?  I'll just give you access to that too.

twcclegg:
Thanks, nuget is the same `twcclegg`, once I found out that updating the new metadata required significant code changes I kinda wanted it done before there was urgent need for any updates in them. It might make sense to move the main fork, perhaps should let others weigh in. The main thing if we did do that would be to get Google to point the link in their readme.

aidanbebbington:
I've added you as an owner to nuget now too.  Yea, we were probably quite behind on the code changes, I have literally only updated the metadata.

andreymir:
Hi! Could you update a _Project Site_ url on the nugt package page https://www.nuget.org/packages/libphonenumber-csharp? Currently, it is pointing to the old bitbucket repository which is confusing, because the repository has not been updated since 2012, but 12 new nuget packages have been published since then.

PR #46, opened 2017-10-31
Simplified automatic generated files

matteofabbri:
Now they are easier to read, modify but most important generated addresses can be stored into database like Mongo without loosing pieces


csharp/PhoneNumbers/MetadataFilter.cs
```diff
@@ -327,25 +243,18 @@ internal void FilterMetadata(PhoneMetadata.Builder metadata)
                 }
             }
             foreach (var wildcardChild in wildcardChildren)
+            foreach (var parent in ExcludableParentFields)
```
twcclegg: This bracket reduction seems a bit extreme

twcclegg (commented):
Looks good! Can you modify or add to one of the readmes on how you massaged the proto generated code so it can be recreated whenever the protos are next updated?
Also thankfully doesn't seem like it will be rebase hell with the work I have outstanding :)

PR #58, opened 2018-03-23
Add support for .NET 3.5

neuroid:
This is more of a suggestion than anything else. I understand that .NET 3.5 at this point is considered legacy and that in the long run supporting it might turn out to be a maintenance burden. However, some might find this useful (i.e. myself :wink:).

There's actually surprisingly little code that had to be changed. The missing `SortedSet<T>` was replaced with a custom implementation, based on the `SortedDictionary<T>` to achieve similar performance. It's not a complete `SortedSet<T>` implementation by any means; it only contains enough functionality to support the current use. This class and the corresponding interface are only included when targeting `net35`.

### DIFF
--- a/csharp/PhoneNumbers/PhoneNumbers.csproj
+++ b/csharp/PhoneNumbers/PhoneNumbers.csproj
@@ -5,7 +5,7 @@
     <AssemblyTitle>libphonenumber-csharp</AssemblyTitle>
     <VersionPrefix>$(APPVEYOR_BUILD_VERSION)</VersionPrefix>
     <Authors>Patrick Mézard;Tom Clegg;Jarrod Alexander;Google;libphonenumber contributors</Authors>
-    <TargetFrameworks>net40;netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net35;net40;netstandard1.3;netstandard2.0</TargetFrameworks>
     <AssemblyName>PhoneNumbers</AssemblyName>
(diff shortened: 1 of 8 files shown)


twcclegg:
The project had 3.5 support until about a year ago when netstandard support was added. The issue isn't lacking libraries but building and packing it https://github.com/Microsoft/msbuild/issues/1333
Seems that there is a fix https://github.com/mj1856/TimeZoneConverter/commit/990bb739b266cbe217c458b8e394684d4bab532b if you want to add that I'll release this with the next metadata update.

csharp/PhoneNumbers/PhoneNumbers.csproj
```diff
@@ -28,4 +28,19 @@
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="compat\ISet.cs" />
+    <Compile Remove="compat\SortedSet.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
```
twcclegg: I don't believe you need this section.

neuroid: Cool. Removed.

neuroid:
I added the fix. Thanks for reviewing!

twcclegg (approved):

twcclegg:
v8.9.3 is now available in nuget with .net35 support. Thanks for the contribution!

PR #63, opened 2018-05-24
Initial Func

dchankhour:


twcclegg:
This library is only to be ports of functionality of the the upstream https://github.com/googlei18n/libphonenumber/tree/master/java other functionality should be added elsewhere.
As for this file, I don't know it's history, it hasn't been modified to track the java version since 2012. It no longer exists in the upstream repo, it seems it was moved to https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/ShortNumberInfo.java I'd start looking at that along with https://github.com/twcclegg/libphonenumber-csharp/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java#L567. Also work to bring this port closer to feature parity should be made to my `geocoding` branch.

PR #75, opened 2018-09-21
Adding Verify as extension method to Leniancy

wallymathieu:
In order to match java implementation:
https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java#L477

### DIFF
--- a/csharp/PhoneNumbers/LeniencyExtensions.cs
+++ b/csharp/PhoneNumbers/LeniencyExtensions.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using Leniency=PhoneNumbers.PhoneNumberUtil.Leniency;
+namespace PhoneNumbers
+{
+    public static class LeniencyExtensions
+    {
+        public static bool Verify(this Leniency leniency, PhoneNumber number, string candidate, PhoneNumberUtil util)
+        {
+            switch (leniency)
+            {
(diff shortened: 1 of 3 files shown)


twcclegg:
Minus the minor build error, I like it; I am however reluctant to merge it. It breaks the API contract which shouldn't happen unless it's reved to 9.0, I've keep the release version pinned to the Google version for simplicity and visibility. I have no idea when they might move to 9.0.

wallymathieu:
Well, perhaps a better way would then be to change this PR so that it doesn't break the contract, instead do a
```c#
    public static class LeniencyExtensions
    {
        public static bool Verify(this Leniency leniency, PhoneNumber number, string candidate, PhoneNumberUtil util)=>util.Verify(leniency, number, candidate, util);
     }
```

twcclegg:
Optionally, if you would like, you could do it the other way around and have the legacy `Verify` call the enum extension and mark it Obsolete. Either way I'll merge this in time to be in the next release unfortunately 8.9.14 came out only a few hours so it might be a week or two.

wallymathieu:
It's OK, I just thought that I would help out a bit while trying to figure things out :wink:.
This is not a prio, more of a nice to have in order to make the API make more sense for the next guy.

PR #76, opened 2018-09-27
Add code example

tinohager:
phone numer is a fake number used on other projects.
https://github.com/giggsey/libphonenumber-for-php/blob/master/docs/PhoneNumberUtil.md


twcclegg:
Thanks!

PR #80, opened 2018-10-16
Use embedded xml resources for .NET 3.5

athoik:
Since 6a75a40218b we are not trying to use dll embedded resources.
Instead we are tying to load directly the XDocument from filesystem.

This commit should fix the problem by using embedded dll resources.
Although further tests required to ensure that external resources still work.

### DIFF
--- a/csharp/PhoneNumbers/BuildMetadataFromXml.cs
+++ b/csharp/PhoneNumbers/BuildMetadataFromXml.cs
@@ -91,7 +91,7 @@ namespace PhoneNumbers
             using (var input = asm.GetManifestResourceStream(name))
             {
 #if NET35
-		document = XDocument.Load(new XmlTextReader(input));
+                document = XDocument.Load(new XmlTextReader(input));
 #else
                 document = XDocument.Load(input);
 #endif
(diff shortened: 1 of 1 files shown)


csharp/PhoneNumbers/BuildMetadataFromXml.cs
```diff
@@ -80,19 +83,19 @@ public class BuildMetadataFromXml
...
-#if  NET40
+#if (NET35 || NET40)
             var asm = Assembly.GetExecutingAssembly();
 #else
             var asm = typeof(PhoneNumberUtil).GetTypeInfo().Assembly;
 #endif
             using (var input = asm.GetManifestResourceStream(name))
             {
+#if NET35
+		document = XDocument.Load(new XmlTextReader(input));
```
twcclegg: nit but, can you fix the whitespace here?
```suggestion
                document = XDocument.Load(new XmlTextReader(input));
```

twcclegg (approved):

csharp/PhoneNumbers/BuildMetadataFromXml.cs
athoik: Sure ;)

Please consider "squash and merge", because all those were trial and error!

PR #117, opened 2020-05-20
Add semicolon to singleExtensionSymbolParsing

sbebrys:
To consistent state with https://github.com/google/libphonenumber/blob/f52af239997c35335894206a4a45f44558ee71a0/cpp/src/phonenumbers/phonenumberutil.cc#L687


twcclegg:
Thanks!

PR #119, opened 2020-07-06
Make 2 degrees mobile numbers valid for NZ

amul047:


amul047:
Hi @twcclegg , what's your release process like? It appears 8.12.5 is not on NuGet yet, so I am assuming it needs some intervention

twcclegg:
The release process is triggered by the github release tag, 8.12.5 not getting released was an oversight. Both 8.12.5 and 8.12.6 are now published however neither of them (nor 8.12.7) appear to have the NZ change, you would need to go through https://github.com/google/libphonenumber/blob/master/CONTRIBUTING.md#filing-a-metadata-issue

amul047:
@twcclegg logged issue report here - https://issuetracker.google.com/issues/160627309

amul047:
Looks like the number was in fact invalid as per ITU documentation. I will abandon this PR

Issue #41, opened 2017-09-26
Where is PhoneNumberToCarrierMapper and PhoneNumberToTimeZonesMapper


swtanggara:
Hi @aidanbebbington,

Thanks for your work on this port of library. I want to ask this because I need the PhoneNumberToCarrierMapper and PhoneNumberToTimeZonesMapper classes, which is not found in this port. Or is there anyone else you guys try to present these classes in this port. If I want add these classes on this repository, can you please guide me to do that.

Thanks.

Danthar:
PhoneNumberToCarrier mapper? What would you expect that to do ? Because numbers change carrier all the time, so i wonder how that would work in a library such as this :)

swtanggara:
Hi @Danthar,

Sorry for my incomplete question. I found these classes (in java) on the main repository of https://github.com/googlei18n/libphonenumber (https://github.com/googlei18n/libphonenumber/blob/master/java/carrier/src/com/google/i18n/phonenumbers/PhoneNumberToCarrierMapper.java and https://github.com/googlei18n/libphonenumber/blob/master/java/geocoder/src/com/google/i18n/phonenumbers/PhoneNumberToTimeZonesMapper.java). 

I assume that this repo not updated or synced with that original repo. That original repo may now have a lot of changes.

Danthar:
Ah i see. Its this comment that explains it: https://github.com/googlei18n/libphonenumber/blob/master/java/carrier/src/com/google/i18n/phonenumbers/PhoneNumberToCarrierMapper.java#L109

Its only valid for regions which do not support number portability. But even then i'd say its very iffy. Having no idea how often this information changes. You'd have to really know what your doing when you are relying on such a static source of information.

swtanggara:
Yeah you're right, this validating phone number stuff is indeed relying on a lot of static information. 

Currently I'm trying to create my own port of libphonenumber on local by examining at two source (original libphonenumber and this libphonenumber-csharp), and to be honest it's hard because my limited expertise both on C# and Java.

Danthar:
Great way to learn :) Just be glad its not scala

twcclegg:
@swtanggara The port is most certainly lacking compared to the source. I've been working to bring it up to parity since I took over keeping it up to date. If you would like I could put this higher up in my to-do list currently have been porting things rather ad hoc.

swtanggara:
[...]

PR #128, opened 2020-12-09
initial carrier work, more immutable, obsolete visiblefortesting

twcclegg:
Addresses #41


Rasmus715:
Is it still possible to fix these check problems?

PR #142, opened 2021-11-10
Add functionality to refresh underlying metadata from a stream.

mhepburn:
The purpose of this change is to allow a new instance of a PhoneNumberUtil to be created with a fresh copy of the metadata to allow dynamic updates for validation that don't require a rebuild/restart of the service using it.  
Thanks!

### DIFF
--- a/csharp/PhoneNumbers/BuildMetadataFromXml.cs
+++ b/csharp/PhoneNumbers/BuildMetadataFromXml.cs
@@ -80,10 +80,10 @@ namespace PhoneNumbers
             => BuildPhoneMetadata(name, null, liteBuild, specialBuild, isShortNumberMetadata, isAlternateFormatsMetadata, nameSuffix: false);
 
         internal static PhoneMetadataCollection BuildPhoneMetadata(string name, Assembly asm = null,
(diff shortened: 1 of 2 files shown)


twcclegg:
What is the purpose of this? Currently there is no way to use metadata other than what is include. In general the intention is for this port to _only_ match the [Java](https://github.com/google/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java) functionality which similarly does not allow for controlling the metadata.
Metadata only updates every 2-3 weeks.

mhepburn:
The plan is to grab the latest metadata.xml file from github and refresh it with that.  

The reason is that I have some services that rarely get updates that do validation with the library.  I don't know when we're going to use new area codes that come out so there's generally a couple days of lag between it being reported and being able to update it.  Instead I'd like to just dynamically load the new .xml once a day if a validation fails.

mhepburn:
I initially saw this createInstance method in the java version that lets you send a custom metadataLoader that is what gave me the idea:
https://github.com/google/libphonenumber/blob/libphonenumber-7.5.2/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java#L988-L1005

Issue #136, opened 2021-07-08
Performance issues


RobThree:
I've been using [IKVM](https://www.ikvm.net/) to "cross-compile" the [libphonenumber jars](https://repo1.maven.org/maven2/com/googlecode/libphonenumber/libphonenumber/) to a .Net DLL for a long time. Unfortunately, IKVM is [no longer maintained](http://weblog.ikvm.net/CommentView.aspx?guid=33ea525f-a291-418a-bd6a-abdf22d0662b) (for quite some while now). Today I replaced the cross-compiled DLL with this library, which was a breeze since the public API is, besides .Net conventions like `getInstance()` -> `GetInstance()` and a single incompatibility with `TruncateTooLongNumber` requiring a builder,  very compatible. All was well, all my unittests passed. 🎉🥳 So thank you, and kudo's for the great work on making this such an easy transition.

However, then I started benchmarking and, it turns out, this lib is slower by a factor 30. When handling a lot of phonenumbers that is quite a lot. I have browsed the issues and found some and I'm quite sure and convinced the culprit is the regex(es) used in this lib. I have tested it on .Net 5.0 in a windows console application.

I wish I could dive in deeper and profile the issue(s) and put in some work to help alleviate this but I'm not in a position right now - too busy. However, _should_ I have some spare time; which (of the many) branch(es) should I base my work on? What would be my best bet? I can't make any promises but I would love to help improve this library.

twcclegg:
Hi there, any support is always appreciated. The regexs almost certainly are the bottleneck, there's been some work to make things faster https://github.com/twcclegg/libphonenumber-csharp/pull/121. Master would be the place for any contribution. Thanks in advance (and also for prompting me to clean up the stale branches).

RobThree:
With the [regex performance improvements](https://devblogs.microsoft.com/dotnet/regular-expression-improvements-in-dotnet-7/) in the [earlier this week released .Net 7](https://devblogs.microsoft.com/dotnet/announcing-dotnet-7/) there could be some real difference. I haven't gotten around to it since july 2021 but I hope to play a little with this and see what the differences are soon.

RobThree:
Ok, so I haven't done anything with the .Net versions (yet) but could improve performance already with about a factor of 10 but I did have to sacrifice the RegexCache's internal LRU. As the commit says:

[...]

PR #161, opened 2022-11-11
Improve (unittest) performance by a factor of ~10 (expected real-world improvements to be even higher)

RobThree:
PhoneRegex now uses Lazy<Regex> and the RegexCache now uses a ConcurrentDictionary internally.

I have sacrificed the LRU cache, which is why the TestRegexInsertion currently fails. This means Regexes are currently cached indefinitely. I wonder in how many real-world scenario's caching a few hundred(?) regexes would be an actual problem. Unfortunately I have no numbers on the use of this class nor any idea about the environments this class is used in. But the added complexity of the LRU (and making it thread-safe) seems not worth it to me. Especially if this easy optimization alone results in a 10x improvement.

My numbers for running all unittests went from ~18seconds to ~1.8 seconds on my 9 year old workstation. Will discuss further in #136

### DIFF
--- a/csharp/PhoneNumbers/PhoneRegex.cs
+++ b/csharp/PhoneNumbers/PhoneRegex.cs
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+using System;
 using System.Text.RegularExpressions;
 
 namespace PhoneNumbers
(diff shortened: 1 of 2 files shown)


RobThree:
[In my tests](https://github.com/twcclegg/libphonenumber-csharp/issues/136#issuecomment-1313909023) with real world data I've seen an performance improvement of about a factor 70 in, at least, parsing the phonenumbers.

twcclegg:
Assuming this is good to go, I'm happy to merge. Thanks!

RobThree:
You may want to check my [remarks here](https://github.com/twcclegg/libphonenumber-csharp/issues/136#issuecomment-1314149456) but, yes, this should be good to go 🚀

RobThree:
Hi @twcclegg 

Was there a problem merging? Or did you just not get around to it? If there's anything I can do to help, please let me know.

twcclegg:
I didn't get notifications of either of the last two messages for some reason :( , so that would be why.
Thanks again.

RobThree:
No worries and thank **you**!

Issue #160, opened 2022-10-10
How to parse without getting exception for wrong format?


ziaulhasanhamim:
I'm getting a phone number input from the user. So I try to parse it using the `Parse` method of `PhoneNumberUtil`. If the number format is incorrect then it will throw an exception. But I don't want to get an exception. I want something like `TryParse` which will return a boolean containing success or failure. Is there an easy to achieve this behavior?

twcclegg:
This port tries to parity Google's Java version and as such doesn't have some patterns more common to C# e.g. TryParse.
Probably the best and simplest would be just wrapping Parse with a try/catch block.

ziaulhasanhamim:
Adding TryParse is no harm after all it's a useful method that most C# developers gonna look for. Like there isn't a need of changing anything in the port itself. TryParse will do the same thing returning false for failure instead of throwing exceptions.

brolaugh:
Adding `TryParse` to the library is a slippery slope to adding more things that isn't strictly 1-to-1 with Google's Java version. 
It will lead to questions like:

> You added the `TryParse` method so why not add the `FooBar` method as well?

ziaulhasanhamim:
> You added the TryParse method so why not add the FooBar method as well?

There is nothing wrong with adding FooBar also if it makes the developer experience better.

But yeah I can understand that thing strict port. Like the API should be one to one with java version

angelaki:
How about just creating an extension method, if you need it?

```
    public static bool TryParse(this PhoneNumbers.PhoneNumberUtil util, string numberToParse, string defaultRegion, out PhoneNumbers.PhoneNumber? phonenumber)
    {
        phonenumber = null;

        try { phonenumber = util.Parse(numberToParse, defaultRegion); return true; }
        catch { return false; }
    }
```

twcclegg:
Would you like a libphonenumber-csharp.extensions package with that functionality? I would need to ponder what other functionality it might contain, but should be straightforward enough to make happen.

ziaulhasanhamim:
That would be great to have.

twcclegg:
Okay I'll look into it, almost all of the effort is on the logistics side, CI and nuget changes; not any actual code changes.

twcclegg:
Should there be `TryParse` and `TryParseValid` maybe?

twcclegg:
[...]

PR #163, opened 2022-11-26
Feature/tzmapper

senecaconsultancy:


twcclegg:
Thanks!! I likely won’t have time to fully review this before Monday, but happy for the contribution to help bring this up to parity.

senecaconsultancy:
Hi, thank you... I hope it works out.
Thanks for creating this repo!

csharp/PhoneNumbers/PhoneNumbers.csproj
```diff
@@ -29,6 +29,8 @@
   <ItemGroup>
     <EmbeddedResource Include="..\..\resources\*.xml" Exclude="..\..\resources\PhoneNumberMetadataForTesting.xml" />
     <EmbeddedResource Include="..\..\resources\geocoding\**" LinkBase="geocoding" />
+    <EmbeddedResource Include="..\..\resources\timezones\map_data.txt" LinkBase="timezones" />
+    <EmbeddedResource Include="..\..\resources\timezones\windowsZones.xml" LinkBase="timezones" />
```
twcclegg: What is this file? I see the comment below but the resources folder is a direct copy of the Google libphonenumber resources [directory ](https://github.com/google/libphonenumber/tree/master/resources) and this would get whipped out.

csharp/PhoneNumbers/TimezoneMapper.cs
```diff
@@ -0,0 +1,177 @@
...
+namespace PhoneNumbers
+{
+    public class TimezoneMapper
+    {
+        private readonly IDictionary<long, string[]> map;
+        private readonly IDictionary<string, List<string[]>> dotnetmap;
+        private readonly ConcurrentDictionary<string, TimeZoneInfo> tziCache;
+        private readonly PhoneNumberUtil phoneUtil;
+
+        internal TimezoneMapper(IDictionary<long, string[]> source, IDictionary<string, List<string[]>> dotnetSource, IList<TimeZoneInfo> initZones)
```
twcclegg: This doesn't seem to match the Google Java API. I was assuming that this was going to be a port of https://github.com/google/libphonenumber/blob/master/java/geocoder/src/com/google/i18n/phonenumbers/PhoneNumberToTimeZonesMapper.java and would at least need to match the API spec.

csharp/PhoneNumbers/PhoneNumbers.csproj
senecaconsultancy: Hi. Okay, I was not aware of the provenance of the directory.  Would it be acceptable to create a new directory that does not exist in the Google libphonenumber? The reason is that it seems more natural for .Net applications to be able to get a TimeZoneInfo object, and the xml file maintained by unicode seems like an easy solution.

However, this part could be omitted if strict conformance to the Google API is required.

csharp/PhoneNumbers/TimezoneMapper.cs
senecaconsultancy: Hi. It's true that I did not implement this with adherence to the Google API.
I can change that if that is the goal of this port and what people want.

As an aside, I didn't implement it that way because returning a list with one element "Etc/Unknown" seems more burdensome on the callers than an empty collection.

If you want me to re-do this with strict conformance, I am willing to do that. Please let me know.
Thanks for your comments.

twcclegg:
Yeah when I inherited the library it was a quite strict port, perhaps maybe even stricter than it is now. But I would still like to maintain all API contracts, be at least somewhat based on Java implementations etc. (to the point that it arguably should parity Google's behavior even if it could be seen as a bug).

csharp/PhoneNumbers/PhoneNumbers.csproj
twcclegg: Per this additional potential functionality I have started work on an idea based on https://github.com/twcclegg/libphonenumber-csharp/issues/160 (libphonenumber-csharp.extensions) it does seem like that could be a good place for this and other OTT functionality. This would certainly add more validity to a side library than just `TryParse`

twcclegg: WIP https://github.com/twcclegg/libphonenumber-csharp/pull/164

senecaconsultancy:
Hi. I think this update is much closer to your expectations.
Please review when possible. Thank you.

csharp/PhoneNumbers.Test/TestRegexCache.cs
```diff
@@ -48,7 +48,7 @@ public void TestRegexInsertion()
             regexCache.GetPatternForRegex(regex3);
             Assert.True(regexCache.ContainsRegex(regex3));
 
-            Assert.False(regexCache.ContainsRegex(regex2));
+            Assert.True(regexCache.ContainsRegex(regex2));
```
twcclegg: You can just nix this line, side effect of ditching the LRU due to performance issues. I found tests failures weren't failing the CI.

csharp/PhoneNumbers/PhoneNumberToTimeZonesMapper.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace PhoneNumbers
+{
+    public class PhoneNumberToTimeZonesMapper
+    {
+        private static readonly string[] UNKNOWN_TIMEZONE = { "Etc/Unknown" };
+
+        private readonly IDictionary<long, string[]> map;
```
twcclegg: Can this be ImmutableDictionary?

csharp/PhoneNumbers/TimezoneMapDataReader.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        /// <summary>
+        /// Consumes 'map_data.txt' from this repository and returns a mapping from numerical prefixes
+        /// of phone numbers to IANA time zone names associated with the number.
+        /// </summary>
+        /// <param name="fp">Input stream for 'map_data.txt'</param>
+        /// <param name="splitters">array of char that delimits separate time zones in a string.</param>
+        /// <returns></returns>
+        public static IDictionary<long, string[]> GetPrefixMap(Stream fp, char[] splitters)
```
twcclegg: This should be able to be `internal`.

twcclegg (approved):
A couple small things but otherwise LGTM.

csharp/PhoneNumbers.Test/TestRegexCache.cs
senecaconsultancy: yes, deleted it.

csharp/PhoneNumbers/TimezoneMapDataReader.cs
senecaconsultancy: excellent point. thanks.
made class and method internal

csharp/PhoneNumbers/PhoneNumbers.csproj
senecaconsultancy: yeah, I think it makes sense to include the dotnet timezone mapping thing in an extension.

csharp/PhoneNumbers/PhoneNumberToTimeZonesMapper.cs
senecaconsultancy: yes, good point.
done.

PR #176, opened 2023-07-14
Cleanup package dependencies

pentp:
Cleanup unnecessary package dependencies and solution configurations.


twcclegg:
Thanks. Not surprised there's cruft around when people still were wanting net35 and mobile.

PR #178, opened 2023-07-19
Reduce regex abuse

pentp:
* Use a single cache for all `PhoneRegex` instances.
* Replace all static `PhoneRegex` instances with regular `Regex` instances.
* Remove some really trivial regexes and use normal C# code instead.
* Use `GeneratedRegex` on .NET 7 for static regexes.
* Optimize many regexes to be simpler and avoid backtracking whenever possible.
* Disable number metadata regex validation in production code.


twcclegg:
Cool! Thanks, I had started on moving to generated regexs here https://github.com/twcclegg/libphonenumber-csharp/pull/175 but that was just a first pass and is unfinished. Looks good, I'll review it more thoroughly tomorrow.

pentp:
Did some benchmarking also:

|                   Baseline |       Mean |    Error |   StdDev | Allocated |
|------------------------- |-----------:|---------:|---------:|----------:|
|                    Parse |   948.5 ns | 18.92 ns | 23.93 ns |     945 B |
| GetCountryCodeFromNumber | 1,606.1 ns |  5.61 ns |  5.25 ns |    1173 B |
|            GetNumberType | 2,686.6 ns |  9.85 ns |  8.74 ns |    1553 B |

|                   This PR|       Mean |    Error |   StdDev | Allocated |
|------------------------- |-----------:|---------:|---------:|----------:|
|                    Parse |   729.7 ns |  8.07 ns |  6.74 ns |     879 B |
| GetCountryCodeFromNumber | 1,385.7 ns |  8.01 ns |  7.10 ns |     909 B |
|            GetNumberType | 2,280.1 ns | 11.65 ns | 10.89 ns |     948 B |

csharp/PhoneNumbers/PhoneNumberUtil.cs
```diff
@@ -123,63 +96,87 @@ public partial class PhoneNumberUtil
...
-        private static readonly Regex CapturingDigitPattern;
-
-        /// <summary>Regular expression of acceptable characters that may start a phone number for the purposes of
-        /// parsing. This allows us to strip away meaningless prefixes to phone numbers that may be
-        /// mistakenly given to us. This consists of digits, the plus symbol and arabic-indic digits. This
-        /// does not contain alpha characters, although they may be used later in the number. It also does
-        /// not include other punctuation, as this will be stripped later during parsing and is of no
-        /// information value when parsing a number.
-        /// </summary>
-        public static readonly Regex ValidStartCharPattern;
```
twcclegg, line 172: I don't see any usage on public GitHub but I'm a little hesitant to remove public items. However it's private in the java source, (there's plenty of things that shouldn't be public but was done before I inherited this project) so it's probably okay enough in this case.

csharp/PhoneNumbers/PhoneRegex.cs
```diff
@@ -15,31 +15,29 @@
...
-        public PhoneRegex(string pattern)
-            : this(pattern, RegexOptions.None)
-        {
-        }
+        private static readonly ConcurrentDictionary<string, PhoneRegex> cache = new();
+
+        public static PhoneRegex Get(string regex) => cache.GetOrAdd(regex, k => new PhoneRegex(k));
 
-        public PhoneRegex(string pattern, RegexOptions options)
+        private PhoneRegex(string pattern)
```
twcclegg: Similar to the above, I don't know if this is ever used directly by a consumer but would be a breaking change and I don't think I want to make a major version bump at this time.

csharp/PhoneNumbers/RegexCache.cs
```diff
@@ -1,42 +0,0 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-using System;
-using System.Collections.Concurrent;
-
-namespace PhoneNumbers
-{
-    public class RegexCache
```
twcclegg, line 22: Same as above.

twcclegg (commented):
All looks good minus the public changes which might be acceptable.
Thanks for all your ongoing contributions. Certainly nice to have someone from Microsoft that really knows regex tuning. 🍻

csharp/PhoneNumbers/PhoneNumberUtil.cs
pentp: Would adding a `[Obsolete(error: true)]` attribute be a reasonable mitigation in this case (with a null value for the actual regex)?

twcclegg: I think that sounds reasonable. With that I can merge this and release (Google's upstream came out about 12 hours ago).
It is kinda breaking, but it's things that shouldn't have ever been public, and not sure what value they'd provide; and I'm sure I'll get comments here or via nuget if it causes any problems.

twcclegg:
Anything remaining on your end, or good to merge?

pentp:
Should be good to merge.

twcclegg:
Thanks again for all your help

PR #180, opened 2023-07-31
Skip builders internally. Reduce allocations.

pentp:
* Skip all builder types internally to reduce both allocation and throughput overhead.
* Reduce metadata/`PhoneNumber` object size by removing `Has*` fields and packing some fields to smaller types.
* Use built-in integer formatting instead of custom code on .NET 6/7.
* Remove many `String(Builder)` allocations.

Benchmark results:

|                   Baseline|       Mean |    Error |   StdDev | Allocated |
|------------------------- |-----------:|---------:|---------:|----------:|
|                    Parse |   729.7 ns |  8.07 ns |  6.74 ns |     879 B |
| GetCountryCodeFromNumber | 1,385.7 ns |  8.01 ns |  7.10 ns |     909 B |
|            GetNumberType | 2,280.1 ns | 11.65 ns | 10.89 ns |     948 B |

|                   This PR|       Mean |   Error |  StdDev | Allocated |
|------------------------- |-----------:|--------:|--------:|----------:|
|                    Parse |   526.2 ns | 0.76 ns | 0.64 ns |     477 B |
| GetCountryCodeFromNumber |   921.8 ns | 5.35 ns | 4.74 ns |     507 B |
|            GetNumberType | 1,485.5 ns | 6.08 ns | 5.39 ns |     546 B |


pentp:
The code coverage diff is mostly because I removed `[CompilerGenerated]` from most of the metadata classes.

twcclegg (approved):

Issue #172, opened 2023-06-15
Make metadata update and publish new version fully automatically


twcclegg:
Currently metadata update is [semi-automated ](https://github.com/twcclegg/libphonenumber-csharp/blob/main/csharp/lib/update.sh). The remaining work is to have this script run automatically, likely triggered via email from libphonenumber-discuss@googlegroups.com "libphonenumber {VERSION} released" or polling https://github.com/google/libphonenumber/releases.

wmundev:
Hey @twcclegg 
i have a PR to resolve this issue up - https://github.com/twcclegg/libphonenumber-csharp/pull/192

let me know what you think, hope it helps!

PR #185, opened 2023-11-04
feat: Create dependabot.yml for automatic updates for nuget packages

wmundev:
This PR will automatically create pull requests whenever a new dependency is available for the dotnet projects, e.g. if xunit released a new package

it checks weekly currently but can be tweaked according to your preference

See example pr raised by dependabot
image
image

hope that helps!

fyi @twcclegg


twcclegg:
Thanks!
What would be really great would be would be a automatic updater for the metadata https://github.com/twcclegg/libphonenumber-csharp/issues/172

PR #190, opened 2023-11-13
feat: fix warnings as error issue with phonenumber unit tests

wmundev:
This should hopefully fix issues with the unit tests not running


twcclegg:
Caught this before I saw this PR, thanks though.

PR #202, opened 2023-11-25
feat: fix tests failing when github actions metadata update script runs on a linux machine

wmundev:
[github action run](https://github.com/twcclegg/libphonenumber-csharp/actions/runs/6987957696) is failing due to the path separator, fix the issue by using both `Path.DirectorySeparatorChar` or `Path.AltDirectorySeparatorChar` and ensure zip file is parsed properly

@twcclegg this should fix the issue with the github action i've created

See successful run in my test project - https://github.com/wmundev/test-test-libphonenumber-csharp/actions/runs/6988539197/job/19016203551

See new release created successfully - https://github.com/wmundev/test-test-libphonenumber-csharp/releases/tag/v8.13.26

### DIFF
--- a/csharp/PhoneNumbers/PhoneNumberOfflineGeocoder.cs
+++ b/csharp/PhoneNumbers/PhoneNumberOfflineGeocoder.cs
@@ -164,13 +164,7 @@ namespace PhoneNumbers
 
         private static string[] ParseNameFromArchive(ZipArchiveEntry entry)
         {
-            const int expectedLength = 2;
-            var name = entry.FullName.Split('.')[0].Split(Path.DirectorySeparatorChar);
-            if (name.Length < expectedLength)
-            {
-                name = entry.FullName.Split('.')[0].Split(Path.AltDirectorySeparatorChar);
-            }
-            return name;
+            return entry.FullName.Split('.')[0].Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
         }
 
         private static SortedDictionary<int, HashSet<string>> LoadFileNamesFromManifestResources(Assembly asm, string prefix)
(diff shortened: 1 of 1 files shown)


csharp/PhoneNumbers/PhoneNumberOfflineGeocoder.cs
```diff
@@ -162,6 +162,17 @@ internal PhoneNumberOfflineGeocoder(string phonePrefixDataDirectory, Assembly as
             return files;
         }
 
+        private static string[] ParseNameFromArchive(ZipArchiveEntry entry)
+        {
+            const int expectedLength = 2;
+            var name = entry.FullName.Split('.')[0].Split(Path.DirectorySeparatorChar);
```
twcclegg: Can you just do a `.Split(Path.DirectorySeparatorChar, Path.AlthDirectorySeparatorChar)`?

twcclegg: Also needs to be verified to also work with the AppVeyor Windows build CI. On the one hand it would be nice to port that job to also use GHAs but building on Windows allows for running unit tests against netframework.

wmundev: does the existing pull request check cover passing on the AppVeyor build? - https://ci.appveyor.com/project/twcclegg/libphonenumber-csharp/builds/48607119

Otherwise, i have also created new github actions to cover both windows and linux - https://github.com/twcclegg/libphonenumber-csharp/pull/204

wmundev: thanks for the suggestion, i've updated the code as recommended in this commit - https://github.com/twcclegg/libphonenumber-csharp/pull/202/commits/499a2fc4e6bbf1cd5eb1571acf07319a719e4dc9

PR #204, opened 2023-11-28
feat: add github action pr builds for building project, and running tests on linux and windows

wmundev:
FYI @twcclegg , these are PR builds to run the unit tests on github actions whenever someone makes a new PR if you are interested in it

github actions are failing because of the zip files delimiter now only supports `\` and not `/`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,12 @@
 
 version: 2
 updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 2
   - package-ecosystem: "nuget"
     directory: "/csharp/PhoneNumbers"
     schedule:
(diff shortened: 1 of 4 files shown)


twcclegg:
Hey there, thanks for all your contributions.
Per the path separator, that seems to have caused some problem https://github.com/twcclegg/libphonenumber-csharp/issues/212
Eventually I would like to fully move all of the CI from AppVeyor to GHA; one consideration being, if it's deemed okay to drop running tests on netframework or if it's worth the added complexity of having to add a Windows build step.

twcclegg:
Given all of your work, any interest in becoming a more official collaborator?

wmundev:
@twcclegg thanks for the link to the issue, i've been trying to replicate the IndexOutOfRangeException issue but to no avail with the code change i made around the path separator, something that might be close that i found is this page [here](https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/mitigation-ziparchiveentry-fullname-path-separator)

however it doesn't make sense to me as the code change we did should have supported both path separators

yeah sure happy to be a official collaborator 👍

wmundev:
@twcclegg what do you think about this implementation of a safe fallback if the original implementation of the directory char doesn't work? - https://github.com/twcclegg/libphonenumber-csharp/pull/204/commits/7d1dccdd0bdf8ab9eb59cf446ea4f9bb8ad2adcc

this will resolve the pr builds failing but also ensure original implementation works as expected

twcclegg (approved):

PR #222, opened 2024-01-02
feat: add editorconfig to this project for standardised code styling

wmundev:
editorconfig adapted from https://github.com/dotnet/runtime/blob/main/.editorconfig

formatted an example dotnet file to show you what it looks like


twcclegg:
Thanks!

PR #234, opened 2024-02-23
feat: initial enable nullable in main project

wmundev:


twcclegg:
Interesting approach, I hadn't looked at disabling certain classes. Given it's all proto based it makes most all of the internal workings a little problematic for nullable. If we're able to get the public interface working that's probably sufficient.

wmundev:
thanks @twcclegg! , i'll give it a go and disable nullable for those internal classes and get the code coverage up. a great benefit of this approach i've found is we can slowly migrate classes to use nullable as needed

wmundev:
@twcclegg nullable update is all working now, build succeeds

is there any classes that you are aware of that i should be disabling instead of modifying in this PR?

added test coverage where possible and left out test coverage where it is not covered in the main google libphonenumber repo

PR #239, opened 2024-03-25
feat: update test projects to dotnet 8

wmundev:


twcclegg:
We can add net8 whenever, but we want to keep net7 until at least [May 14](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core).

wmundev:
thanks Thomas, i'll restore dotnet 7

twcclegg (approved):

PR #264, opened 2024-08-23
Target .NET 8.0 and drop .NET 7.0

filipw:
Hello 👋 This PR updates the targeting of the library.

## Changes
- .NET 8.0 is now explicitly targeted by the "main" library. It was already used in tests, but the main one only targeted 6, 7 and .NET Standard 2.0. 
- .NET 7.0 was removed as it [reached the end of life](https://devblogs.microsoft.com/dotnet/dotnet-7-end-of-support/) in May 2024

Without this, using the library on .NET 8.0 means going over .NET Standard 2.0 which in turn pulls a ton of very old (dating back to 2016) dependencies via [System.IO.Compression](https://www.nuget.org/packages/System.IO.Compression#dependencies-body-tab), some of which even have [CVEs](https://github.com/advisories/GHSA-5f2m-466j-3848) on them.

### DIFF
--- a/.github/workflows/build_and_run_unit_tests_linux.yml
+++ b/.github/workflows/build_and_run_unit_tests_linux.yml
@@ -25,6 +25,6 @@ jobs:
       - name: Build solution
         run: dotnet build --no-restore
         working-directory: ./csharp
-      - name: Test solution targeting dotnet7.0 only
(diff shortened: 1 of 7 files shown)


twcclegg:
Thanks!

PR #273, opened 2024-11-13
.net 9 migration

twcclegg:
## Changes
-


wmundev (approved):
does this need updating too? https://github.com/twcclegg/libphonenumber-csharp/blob/main/lib/github-actions-metadata-update.sh#L116

otherwise looks good to me, nice one!

PR #288, opened 2025-03-05
Add strong name

twcclegg:
## Changes
-

### DIFF
--- a/.github/workflows/build_and_run_unit_tests_linux.yml
+++ b/.github/workflows/build_and_run_unit_tests_linux.yml
@@ -15,6 +15,9 @@ jobs:
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: 9.x
+      - name: Add snk for strong name signing
+        run: |
+          echo "${{ secrets.KEY_SNK_B64 }}" | base64 --decode > key.snk
       - name: Add zip files required for running tests
         run: |
           (cd resources/geocoding; zip -r ../../resources/geocoding.zip *)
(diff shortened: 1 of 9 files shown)


twcclegg:
To be merged before v9.0.0 is released.

wmundev:
@twcclegg have you tried generating an snk key instead? might help with the build errors - https://learn.microsoft.com/en-us/dotnet/standard/assembly/create-public-private-key-pair#create-a-key-pair

then you can use visual studio to sign it - https://learn.microsoft.com/en-us/dotnet/standard/assembly/sign-strong-name#create-and-sign-an-assembly-with-a-strong-name-by-using-visual-studio

twcclegg:
Hey Wilson, yeah that’s my general plan. I have to borrow a Windows machine but should be able to do that in the next couple days.

twcclegg:
~@wmundev I have a feeling that this will break PRs from forks (secrets not available), but we can fix that at a later date.~
This should be resolved now.

wmundev:
@twcclegg nice work on the strong naming!

wmundev:
@twcclegg do you want to consider committing the private key directly to the repo instead of having it in github action secrets? 

this is the recommendation by microsoft - https://learn.microsoft.com/en-us/dotnet/standard/assembly/strong-named#what-makes-a-strong-named-assembly

> Do not rely on strong names for security. They provide a unique identity only.

> If you are an open-source developer and you want the identity benefits of a strong-named assembly for better compatibility with .NET Framework, consider checking in the private key associated with an assembly to your source control system.

twcclegg:
Yeah I saw the thing about just storing it in plain text, I'll likely change it to that. First I'd like to resolve #289 which seems to only effect net framework from what I can tell.
https://github.com/dotnet/runtime/blob/main/docs/project/public-signing.md

PR #297, opened 2025-06-28
feat: fix failing tests by upgrading System.Collections.Immutable

wmundev:
upgrading System.Collections.Immutable seems to have fix the unit tests failing, interesting

@twcclegg do you think its a good idea to upgrade or leave it at 1.7.1?

## Changes
- feat: fix failing tests by upgrading System.Collections.Immutable


twcclegg:
Since the tests are passing it should be fine. Odd that it randomly broke.

twcclegg (approved):

wmundev:
cool, thanks for reviewing, i'll merge it in

PR #322, opened 2026-05-03
fix: port behavior fixes from upstream Java libphonenumber

twcclegg:
- FormatNumberForMobileDialing: add parens around (MX || CL || UZ) so isFixedLineOrMobile applies to all three regions; previously C# operator precedence caused MX/CL to skip the constraint.
- FormatOutOfCountryKeepingAlphaChars: strip any extension already present in the raw input before appending the formatted extension, preventing duplicate extensions in the output (Java #3724).
- FormatInOriginalFormat: call ChooseFormattingPatternForNumber once and reuse the result instead of recomputing it (Java #3976).
- Drop redundant double-assignment in the BR carrier-code branch.

## Changes
-


wmundev (approved):
LGTM

PR #317, opened 2026-05-02
feat: add missing PhoneNumberToCarrierMapper feature that is available in java version

wmundev:
## Changes
- feat: add missing PhoneNumberToCarrierMapper feature that is available in java version

PR #323, opened 2026-05-03
feat: build metadata as per-region binaries; drop runtime XML/zip parsing

twcclegg:
## Summary

Replaces runtime XML metadata parsing + the `geocoding.zip` blob with build‑time-generated per‑region binary files, mirroring the Java upstream's `BuildMetadataProtoFromXml` pipeline.

- New `csharp/PhoneNumbers.MetadataBuilder/` build‑time tool converts `resources/*.xml`, `resources/geocoding/<lang>/<countryCode>.txt`, and `resources/timezones/map_data.txt` into 766 binary files embedded under `PhoneNumbers.metadata.*`, `PhoneNumbers.geocoding.*`, and `PhoneNumbers.timezones.*` resource namespaces.
- `BuildMetadataFromBin` (PhoneMetadata serializer) and `BuildPrefixMapFromBin` (area‑code + timezone serializers) handle read/write; both versioned with magic + format byte.
- New public `IMetadataLoader` interface + `EmbeddedResourceMetadataLoader` default impl + `MissingMetadataException` exception type — consumers can now plug in custom loaders for trimmed bundles, CDN, etc. Mirrors Java's `MetadataLoader` API.
- Lazy `MetadataSource` replaces the eager XML parse on `PhoneNumberUtil.GetInstance()`. Per‑region metadata is loaded only when first asked for, cached in a `ConcurrentDictionary`.
- Geocoding/timezones load via the new binary readers; zip handling and text fallback both removed from `PhoneNumberOfflineGeocoder` / `PhoneNumberToTimeZonesMapper`.

## Outcome

- nupkg: 7.5 MB → 6.9 MB (≈ 8 % smaller).
- No XML parsing on cold start of `PhoneNumberUtil.GetInstance()`.
- Region metadata loaded only on first use (lazy), via a pluggable loader.

## Public API changes — additive only

Verified via diff against `main`: this PR only ADDS public types/members; nothing is removed or signature‑changed. Safe for a non‑major version bump.

**Added:**
- `IMetadataLoader` (interface) + `EmbeddedResourceMetadataLoader` (default impl) — pluggable metadata loading, mirrors Java's `MetadataLoader`.
- `MissingMetadataException` — thrown when an expected metadata resource is missing; mirrors Java's same‑named type.
- `BuildMetadataFromBin`, `BuildPrefixMapFromBin` — read/write helpers for the binary format.
- `MetadataManager.SetMetadataLoader(IMetadataLoader)` — same injection point Java exposes via `DefaultMetadataDependenciesProvider`.

**Behavior‑only correction (no signature change):**
- `PhoneNumberUtil.GetSupportedGlobalNetworkCallingCodes()` returns the same `Dictionary<int, PhoneMetadata>.KeyCollection` as before, but the underlying dictionary now correctly contains only non‑geographical entries (calling codes 800, 808, 870, 878, 881, 882, 883, 888, 979) — matching the docstring and Java's `getSupportedGlobalNetworkCallingCodes()`. The pre‑PR implementation incidentally included every loaded calling code (geo + non‑geo) because it leaked an internal dictionary populated by the eager XML loader; that contradicted the documented contract.

**Removed (internal/non‑public):**
- `AreaCodeParser` (was `internal class`) — replaced by `BuildPrefixMapFromBin.ReadAreaCodeMap`.
- `geocoding.zip` and `testgeocoding.zip` removed from `resources/`; `appveyor.yml` and the GitHub Actions workflows no longer create them.
- `MappingFileProvider.GetFileName` no longer appends `.txt` (internal API; only consumer is `PhoneNumberOfflineGeocoder`).

## Test plan

- [x] PhoneNumbers test suite (322 tests) green on net9.0
- [x] PhoneNumbers.Extensions test suite (23 tests) green on net9.0
- [x] Roundtrip test (`TestBuildMetadataFromBin`) verifies binary I/O against every region in production XML — 200+ phone, 240+ short, 40+ alternate
- [x] Lazy region loading verified by `TestMetadataSource.DefaultLoaderResolvesEmbeddedBinaryMetadata`
- [x] Custom loader injection verified by `TestMedataManager.SetMetadataLoader_RoutesLookupsThroughCustomLoader`
- [x] 10× sequential cold‑from‑empty `dotnet build PhoneNumbers.sln` — 10/10 pass after the mutex + double‑checked‑skip in MetadataBuilder
- [x] Clean build across all four TFMs: netstandard2.0, net8.0, net9.0, netframework4.8

🤖 Generated with [Claude Code](https://claude.com/claude-code)


csharp/PhoneNumbers.Test/PhoneNumbers.Test.csproj
```diff
@@ -2,6 +2,12 @@
 
   <PropertyGroup>
     <TargetFrameworks>netframework4.8;net8.0;net9.0</TargetFrameworks>
+    <!--
+      Pin the C# language version so all TFMs (including netframework4.8, which the SDK
+      otherwise defaults to C# 7.3) accept post-7.3 syntax used by tests: `using var`
+      declarations, `?` nullable annotations under `#nullable enable`, etc.
+    -->
+    <LangVersion>preview</LangVersion>
```
wmundev, line 10:
```suggestion
    <LangVersion>13.0</LangVersion>
```

would this be safer?

wmundev (approved):
generally seems ok, i assume you considered the Google.Protobuf library and it wasn't suited to the needs of this repo? 

perhaps we can also consider if we can pack the package locally and test it before merging this change since it changes the core logic quite a bit

looking forward to the speed increase and size decrease, nice one

wmundev:
@twcclegg also something to note, it looks like the performance tests is saying that this change causes the parse validate format phone number to take longer

this branch - https://github.com/twcclegg/libphonenumber-csharp/actions/runs/25299196033/job/74163036453?pr=323

| Method                             | Job                | Runtime            | PhoneNumberCount | Mean       | Error     | StdDev    | Gen0       | Gen1      | Allocated    |
|----------------------------------- |------------------- |------------------- |----------------- |-----------:|----------:|----------:|-----------:|----------:|-------------:|
| ParseValidateAndFormatPhoneNumbers | .NET 8.0           | .NET 8.0           | 1000             |   3.267 ms | 0.0613 ms | 0.0797 ms |    31.2500 |         - |   1000.56 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 9.0           | .NET 9.0           | 1000             |   3.097 ms | 0.0380 ms | 0.0337 ms |    31.2500 |         - |   1000.56 KB |
| ParseValidateAndFormatPhoneNumbers | .NET Framework 4.8 | .NET Framework 4.8 | 1000             |   7.309 ms | 0.0495 ms | 0.0463 ms |   296.8750 |   39.0625 |   1833.02 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 8.0           | .NET 8.0           | 10000            |  31.840 ms | 0.3560 ms | 0.3156 ms |   375.0000 |         - |  10088.49 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 9.0           | .NET 9.0           | 10000            |  30.656 ms | 0.1785 ms | 0.1582 ms |   406.2500 |         - |  10088.48 KB |
| ParseValidateAndFormatPhoneNumbers | .NET Framework 4.8 | .NET Framework 4.8 | 10000            |  72.720 ms | 1.1924 ms | 1.1154 ms |  2857.1429 |  285.7143 |  18424.22 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 8.0           | .NET 8.0           | 100000           | 321.136 ms | 6.2145 ms | 6.3818 ms |  4000.0000 |         - |  101018.2 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 9.0           | .NET 9.0           | 100000           | 310.182 ms | 5.6781 ms | 5.3113 ms |  4000.0000 |         - |  101018.2 KB |
| ParseValidateAndFormatPhoneNumbers | .NET Framework 4.8 | .NET Framework 4.8 | 100000           | 731.482 ms | 9.4639 ms | 8.8526 ms | 29000.0000 | 4000.0000 | 184226.34 KB |

another branch - https://github.com/twcclegg/libphonenumber-csharp/actions/runs/25308482592/job/74189668638?pr=317

| Method                             | Job                | Runtime            | PhoneNumberCount | Mean       | Error     | StdDev    | Gen0       | Gen1      | Allocated    |
|----------------------------------- |------------------- |------------------- |----------------- |-----------:|----------:|----------:|-----------:|----------:|-------------:|
| ParseValidateAndFormatPhoneNumbers | .NET 8.0           | .NET 8.0           | 1000             |   2.988 ms | 0.0072 ms | 0.0064 ms |    31.2500 |         - |    697.22 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 9.0           | .NET 9.0           | 1000             |   2.895 ms | 0.0233 ms | 0.0206 ms |    31.2500 |         - |    697.22 KB |
| ParseValidateAndFormatPhoneNumbers | .NET Framework 4.8 | .NET Framework 4.8 | 1000             |   6.499 ms | 0.0143 ms | 0.0126 ms |   242.1875 |   31.2500 |   1521.21 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 8.0           | .NET 8.0           | 10000            |  29.896 ms | 0.0839 ms | 0.0701 ms |   406.2500 |         - |   6972.25 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 9.0           | .NET 9.0           | 10000            |  29.035 ms | 0.0571 ms | 0.0506 ms |   406.2500 |         - |   6972.25 KB |
| ParseValidateAndFormatPhoneNumbers | .NET Framework 4.8 | .NET Framework 4.8 | 10000            |  65.891 ms | 0.0771 ms | 0.0683 ms |  2375.0000 |  250.0000 |  15212.16 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 8.0           | .NET 8.0           | 100000           | 299.887 ms | 1.0409 ms | 0.9737 ms |  4000.0000 |         - |  69768.43 KB |
| ParseValidateAndFormatPhoneNumbers | .NET 9.0           | .NET 9.0           | 100000           | 290.297 ms | 1.0893 ms | 0.9096 ms |  4000.0000 |         - |  69768.43 KB |
| ParseValidateAndFormatPhoneNumbers | .NET Framework 4.8 | .NET Framework 4.8 | 100000           | 658.040 ms | 2.4372 ms | 2.1605 ms | 24000.0000 | 3000.0000 | 152137.84 KB |

twcclegg:
## Performance: before vs after

I couldn't get the BenchmarkDotNet `PhoneNumberWorkflowBenchmark` to actually run locally on macOS (BDN's `[SimpleJob(Net48), SimpleJob(Net80), SimpleJob(Net90)]` decorator filters are tricky when the .NET 8 runtime isn't installed locally — it kept exiting at validation), so I instrumented two equivalent workloads directly with `Stopwatch` and ran each 10 times against `PhoneNumbers.dll` from `main` and from this PR head. CI's perf workflow on Windows passed; these numbers complement that with the cold-start measurement that the existing benchmark hides inside `[GlobalSetup]`.

Hardware: M-series Mac, Release build, .NET 9.0, fresh process per measurement.

### Cold start (`PhoneNumberUtil.GetInstance()` + first `Parse`)

This is where the PR moves the needle — the existing benchmark wraps `GetInstance` in `[GlobalSetup]` and so doesn't measure it.

| Phase | `main` (XML) | this PR (per-region binaries, lazy) | delta |
| --- | --- | --- | --- |
| `GetInstance()` (singleton init) | **38.6 ms** (median of 10) | **6.2 ms** | **−84 % / 6.2× faster** |
| First `Parse("+14155552671", "US")` | 4.4 ms | 5.6 ms | +27 % (slower) |
| Total cold path to first parsed number | **~43 ms** | **~12 ms** | **−72 % / 3.6× faster** |

The first-parse regression is expected and intentional: on `main`, `GetInstance` eager-parsed the entire XML metadata so US was already in memory. On the PR, US metadata is lazy-loaded on demand the first time it's requested — that ~1 ms shows up in the parse instead of the constructor. Subsequent parses against the same region pay nothing.

### Steady-state throughput (Parse + IsValid + Format on 20 representative numbers, 100k iterations after 1k warmup)

| | `main` | this PR | delta |
| --- | --- | --- | --- |
| µs / op (median) | 5.16 | 5.13 | within noise |
| ops / sec (median) | 193,728 | 195,354 | within noise |

Per-call code paths in `Parse`, `IsValidNumber`, and `Format` are unchanged by this PR (the diff only touches metadata loading, not the operations that consume it), so the equivalence is expected. Both branches' run-to-run variance was ~3 % per the 10 samples; the inter-branch difference is well inside that.

> Note: my probe uses reflection (`MethodInfo.Invoke`) for the inner loop because it loads `PhoneNumbers.dll` dynamically to compare two builds. The reflection overhead is identical on both branches, so the *comparison* is fair, but the absolute µs/op is higher than `BenchmarkDotNet` would report. The claim here is "no regression," not "this is the achievable peak throughput."

### Package size (re-confirming earlier number)

| | `main` | this PR | delta |
| --- | --- | --- | --- |
| `libphonenumber-csharp.nupkg` | 7.48 MB | 6.90 MB | −0.59 MB (−8 %) |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

twcclegg:
Thanks for the perf catch — that regression is real, and your numbers were the smoking gun I missed. Pushed `e50c423e`:

**Root cause**: `PhoneNumberUtil.GetMetadataForRegion` is called several times per `Parse`, and I had it routed through `MetadataSource.GetMetadataForRegion` which used `ConcurrentDictionary.GetOrAdd(key, Func<K,V>)`. That overload allocates a closure for the lambda on **every** call, including cache hits — which is the common path. So every parse leaked ~5 closures, which lines up with the +45 % allocation delta you observed (~310 B/iter ≈ 5 × ~60 B closure).

**Fix**: `TryGetValue` fast path before `GetOrAdd`. Cache hits now do a lock‑free read with zero allocation; cache misses (rare, once per region) still pay the closure cost, which is fine.

```csharp
public PhoneMetadata? GetMetadataForRegion(string regionCode)
{
    if (regionCache.TryGetValue(regionCode, out var cached))
        return cached;
    return regionCache.GetOrAdd(regionCode, key => Load(key));
}
```

My local stopwatch probe was too noisy and reflection‑overhead‑dominated to detect a 5 % shift, which is why I missed it — the BDN run on Windows CI is what'll confirm. Watching the next run.

> i assume you considered the Google.Protobuf library and it wasn't suited to the needs of this repo?

Considered, didn't go that route. Two reasons:
1. The csharp/README.md "Known Issues" section says the original Patrick Mézard port tried protobuf early and couldn't make it work cleanly with this codebase, and that's why the C# port reads XML at runtime. Switching to Google.Protobuf now would mean re-deriving the wire format from the upstream `phonemetadata.proto`, regenerating C# stubs, then making `PhoneMetadata`/`PhoneNumberDesc`/`NumberFormat` either *be* protobuf messages or wrap them. That's a much bigger surface change than the binary format I added.
2. For a write-once-at-build-time / read-many-at-runtime data file, a custom `BinaryWriter`/`Reader` format is simpler — versioned magic + presence bitmask + length‑prefixed strings. No external dependency, no codegen step, no .proto file to keep in sync with upstream Java schema changes. The format is ~150 lines including comments.

I considered protobuf wire compatibility (i.e. shipping the Java-emitted `PhoneNumberMetadataProto_TW` files verbatim) but Java uses `Externalizable` (Java's serialization), not protobuf, for these specific files. So even matching upstream wouldn't have helped.

> perhaps we can also consider if we can pack the package locally and test it before merging this change since it changes the core logic quite a bit

Strong agree. Locally `dotnet pack -c Release csharp/PhoneNumbers` produces a 6.9 MB nupkg. A quick smoke test plan:
1. Pack from this branch.
2. Reference the local `.nupkg` from a small console app.
3. Verify `PhoneNumberUtil.GetInstance().Parse("+14155552671", "US")` works, and that `MetadataManager.GetAlternateFormatsForCountry(49)` returns non-null Germany data.
4. Spot-check `PhoneNumberOfflineGeocoder.GetInstance().GetDescriptionForNumber(...)` (covers the geocoding-bin path).


Re: the LangVersion suggestion — pushed in the same commit. `13.0` over `preview`.

twcclegg:
## Local nupkg smoke test

Per the suggestion in [the previous comment](https://github.com/twcclegg/libphonenumber-csharp/pull/323#issuecomment-4372663155), packed the PR branch and pointed a fresh consumer project at the resulting `.nupkg` (not the source tree — DLL hashes verified to match between `~/.nuget/packages/libphonenumber-csharp/1.0.0/lib/net9.0/PhoneNumbers.dll` and the consumer's `bin/Release/net9.0/PhoneNumbers.dll`).

### nupkg shape

```
$ dotnet pack csharp/PhoneNumbers -c Release -o /tmp/local-nupkg
Successfully created package 'libphonenumber-csharp.1.0.0.nupkg' (6.9 MB)

$ unzip -l libphonenumber-csharp.1.0.0.nupkg
  lib/netstandard2.0/PhoneNumbers.dll   (10 MB, 766 embedded resources)
  lib/net8.0/PhoneNumbers.dll
  lib/net9.0/PhoneNumbers.dll
  +.nuspec, README.md, icon.png

# verify XML metadata + zips not in package:
$ find inspect -name "PhoneNumberMetadata.xml" -o -name "geocoding.zip" -o -name "testgeocoding.zip"
(empty)

# embedded resource namespaces:
$ strings lib/net9.0/PhoneNumbers.dll | grep "PhoneNumbers\.\(metadata\|geocoding\|timezones\)\." | wc -l
766
```

### Consumer test results (22/22 pass)

```
[PASS] PhoneNumberUtil.GetInstance() returns non-null
[PASS] US region metadata loads
[PASS] US country code == 1
[PASS] TW region metadata loads
[PASS] TW country code == 886
[PASS] Parse +14155552671 succeeded — cc=1 nn=4155552671
[PASS] IsValidNumber(+14155552671) == true
[PASS] Format INTERNATIONAL == "+1 415-555-2671"
[PASS] Format NATIONAL == "(415) 555-2671"
[PASS] AsYouTypeFormatter produces "(415) 555-2671"
[PASS] Non-geo metadata for cc=800 loads
[PASS] GetSupportedGlobalNetworkCallingCodes returns Dictionary.KeyCollection  ← API-stable per @twcclegg's earlier feedback
[PASS] GetSupportedGlobalNetworkCallingCodes contains 800
[PASS] GetSupportedGlobalNetworkCallingCodes does NOT contain US's CC=1 (geo)  ← semantic fix: pre-PR included geo codes
[PASS] MetadataManager.GetAlternateFormatsForCountry(49) — Germany — non-null  (41 number formats)
[PASS] Unknown country code returns null gracefully
[PASS] Geocoder description for +1 650-253-0000 → "Mountain View, CA"
[PASS] Geocoder en for KR (+82 22-1234-567) → "Seoul"
[PASS] Geocoder ko for KR → "서울"
[PASS] Timezones for +14155552671 → [America/Los_Angeles]
[PASS] After SetMetadataLoader(fake), GetAlternateFormatsForCountry returns null  ← new IMetadataLoader injection works
[PASS] Fake loader saw a request for PhoneNumberAlternateFormats_49

All checks passed.
```

Coverage:

| Area | Coverage |
| --- | --- |
| Lazy region metadata loading | US, TW |
| Eager non-geo metadata + KeyCollection API | cc=800 + the API-stability assertion |
| Per-call work (`Parse`/`IsValid`/`Format`) | full round-trip |
| `AsYouTypeFormatter` (separate metadata consumer) | typed-input formatting |
| `MetadataManager` alternate formats | DE positive + bogus-cc null |
| `PhoneNumberOfflineGeocoder` (binary geocoding loader) | EN + KO descriptions |
| `PhoneNumberToTimeZonesMapper` (binary timezone loader) | tz lookup |
| `IMetadataLoader` injection | `SetMetadataLoader` round-trip |

Repro is in [a gist](#) — er, this one's just a single `Program.cs` + `consumer.csproj` referencing `libphonenumber-csharp 1.0.0` from a local NuGet feed at `/tmp/local-nupkg`. Happy to commit it under `csharp/PhoneNumbers.SmokeTest/` if you'd want it as a permanent gate (run as a step after `dotnet pack` in AppVeyor).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

shkarface:
Hello @twcclegg , I'm looking into the source code and somewhat cannot find the equivalent to the PhoneNumberToCarrierMapper function in this port, I'm I missing something or is this not available in the public API?

Edit: I just saw this PR that adds the missing feature: https://github.com/twcclegg/libphonenumber-csharp/pull/317

PR #353, opened 2026-06-16
Security and supply-chain hardening: XXE fix, CI hardening, OpenSSF Scorecard, deterministic builds, and governance docs

twcclegg:
## Changes
-


wmundev (approved):
looks good, nice!

PR #359, opened 2026-07-21
feat: add opt-in java/proto check overrides to the metadata update script

wmundev:
## Changes
- add opt-in java/proto check overrides to the metadata update script

dry run here > https://github.com/twcclegg/libphonenumber-csharp/actions/runs/29901782863

### DIFF
--- a/.github/workflows/create_new_release_on_new_metadata_update.yml
+++ b/.github/workflows/create_new_release_on_new_metadata_update.yml
@@ -6,6 +6,15 @@ on:
   schedule:
     - cron: '0 8 * * *'
   workflow_dispatch:
+    inputs:
+      skip_java_check:
+        description: 'Continue even when the upstream release changes .java files (only tick this once those changes have been reviewed as not needing a port)'
+        type: boolean
+        default: false
+      skip_proto_check:
+        description: 'Continue even when the upstream release changes .proto files'
+        type: boolean
+        default: false
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


lib/github-actions-metadata-update.sh
```diff
@@ -1,125 +1,384 @@
...
 getLatestGitHubRelease() {
-    curl --fail --silent --show-error --location "https://api.github.com/repos/$1/releases/latest" | jq -r .tag_name
+    ghApi "https://api.github.com/repos/$1/releases/latest" | jq -er '.tag_name'
 }
 
 getLatestNugetRelease() {
-    curl --fail --silent --show-error --location "https://www.nuget.org/packages/$1/" | grep 'og:title' | sed "s/.*$1 \([^\"]*\).*/\1/"
+    curl --fail --silent --show-error --location --retry 3 --retry-delay 5 \
+        "https://api.nuget.org/v3-flatcontainer/${1,,}/index.json" \
+        | jq -er '[.versions[] | select(test("-") | not)] | last'
```
twcclegg: `last` assumes the NuGet versions array is ascending.
`[.versions[] | select(...)] | last` trusts the flat-container index.json ordering. In practice it's ascending, but the API spec doesn't strongly guarantee it. Cheap insurance: pipe non-prerelease versions through a version sort and take the max, or `sort -V | tail -1`, rather than relying on document order.

.github/workflows/create_new_release_on_new_metadata_update.yml
```diff
@@ -20,14 +33,27 @@ jobs:
...
+      - name: Setup Java
+        uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95 # v5.6.0
+        with:
+          distribution: temurin
+          java-version: '17'
       - name: Get new metadata information and create new GitHub release
         timeout-minutes: 30
         run: |
-          bash lib/github-actions-metadata-update.sh ${GITHUB_TOKEN}
+          bash lib/github-actions-metadata-update.sh "${GITHUB_TOKEN}"
```
twcclegg: Token here is redundant since it's passed via env var below and technically puts the secret on the process command line (visible to `ps` on the runner).

lib/github-actions-metadata-update.sh
```diff
@@ -1,125 +1,384 @@
...
+    echo "error: $*" >&2
+    if [ -n "${GITHUB_STEP_SUMMARY:-}" ]
+    then
+        echo "$*" >> "${GITHUB_STEP_SUMMARY}"
+    fi
+    exit "${code}"
+}
+
+isTrue() {
+    case "${1,,}" in
```
twcclegg: `${1,,}` is a bash 4.0+ feature. Can do `$(toLower "$1")`

lib/github-actions-metadata-update.sh
```diff
@@ -1,125 +1,384 @@
...
+
 getLatestGitHubRelease() {
-    curl --fail --silent --show-error --location "https://api.github.com/repos/$1/releases/latest" | jq -r .tag_name
+    ghApi "https://api.github.com/repos/$1/releases/latest" | jq -er '.tag_name'
 }
 
 getLatestNugetRelease() {
-    curl --fail --silent --show-error --location "https://www.nuget.org/packages/$1/" | grep 'og:title' | sed "s/.*$1 \([^\"]*\).*/\1/"
+    curl --fail --silent --show-error --location --retry 3 --retry-delay 5 \
+        "https://api.nuget.org/v3-flatcontainer/${1,,}/index.json" \
```
twcclegg: Same as above, can do `${packageId}`

twcclegg:
Thanks for this!

lib/github-actions-metadata-update.sh
wmundev: thanks, addressed

.github/workflows/create_new_release_on_new_metadata_update.yml
wmundev: thanks, addressed by removing it

lib/github-actions-metadata-update.sh
wmundev: thanks, addressed

lib/github-actions-metadata-update.sh
wmundev: thanks, addressed

wmundev:
> Thanks for this!

no worries, thanks for the review! addressed the pr comments

twcclegg (approved):

PR #368, opened 2026-08-02
Remove AppVeyor

twcclegg:
- Removes appveyor.yml, its CODEOWNERS entry, README badge, and the dead lib/update.sh
- Adds publish_nuget.yml — tag-triggered pack + push via nuget trusted publishing (OIDC, no secret)
- VersionPrefix moves from $(APPVEYOR_BUILD_VERSION) to a 0.0.0 placeholder the workflow overrides; drops the $APPVEYOR_REPO_COMMIT literal that was being written into published nuspecs
- Coverage + perf jobs move Windows → ubuntu-24.04-arm; perf workflow renamed, benchmark cache re-keyed
- netframework4.8 scoped to Windows hosts — no CI covers it any more, the main tradeoff
- System.Text.RegularExpressions scoped to net48 (NU1510 once pruning activates)
- Metadata script dispatches publish_nuget explicitly, since GITHUB_TOKEN-authored tags don't trigger workflow runs


wmundev (approved):
looks good, only nit thing is that libphonenumber-csharp\csharp\PhoneNumbers.PerformanceTest\README.md has the renamed file run_performance_tests_windows.yml link in there that needs to be updated

PR #370, opened 2026-08-03
Modernization: stack-overflow fix, frozen collections, drop net9.0, centralized build config

twcclegg:
Thirteen commits, worst-first. Measured on the parse path: **~2% faster**
(`ParseValidateAndFormat` −1.6% at 1000 numbers, −2.6% at 10000, both sides measured on the same
runner), costing **+18.25 KB** one-time allocation per `PhoneNumberUtil` with per-parse allocation
unchanged.

### Correctness

**`fix: stop oversized input overflowing the stack in Normalize`** — `Normalize`,
`NormalizeDigitsOnly`, `NormalizeDiallableCharsOnly` and `ConvertAlphaCharactersInNumber` did
`stackalloc char[number.Length]` on caller-supplied strings with no length guard, so a large input
killed the process with an uncatchable `StackOverflowException`. `Parse` has
`MAX_INPUT_STRING_LENGTH = 250`; these four public statics had nothing. Buffers over 256 chars now
come from `ArrayPool`. The buffer stays sliced to exactly `number.Length` — the normalizers rely on
running out of room to drop characters whose numeric value needs more digits than the character it
came from. netstandard2.0 uses `StringBuilder` and was never affected.

**`fix: return a copy from GetSupportedRegions and freeze the region sets`** — the getter returned
the instance's own `HashSet<string>`, and `PhoneNumberUtil` is a process-wide singleton, so any
caller could `Remove()` a region and break validation for the rest of the process. Java returns
`Collections.unmodifiableSet` here. The return type can't change without a binary break (and
`IReadOnlySet` doesn't exist on netstandard2.0), so it returns a copy; the method has no internal
callers. That also unblocked freezing `supportedRegions` and `nanpaRegions`.

**`fix: stop parallel inner builds racing on clean of the generated bins`** — `dotnet clean` on a
cross-targeting project dispatches Clean to each inner TFM build, so three `RemoveDir` calls raced
on one shared `obj/metadata` and failed with MSB3231. Pre-existing; nothing in CI ran
`dotnet clean` until the determinism step below. Both clean targets now run on the outer build only.

**`fix: copy the region code map on all targets and reject unknown normalize modes`** — self-review
fixes: the frozen-map change had made net8+ copy the constructor's dictionary while netstandard2.0
kept aliasing it, and the mode switch would have silently returned `""` for an unhandled case.

### Performance

**`perf: use frozen collections for the short number and timezone lookups`** — `FrozenSet` for
`ShortNumbersRegionCodeSet` and `FrozenDictionary` for the timezone map on net8+, plus `ContainsKey`
+ indexer collapsed to one `TryGetValue` in `LookUpPrefix`. That loop runs once per digit, so it was
up to ~24 probes of an `ImmutableDictionary` per query; the `TryGetValue` half helps every TFM.

**`perf: freeze the country calling code to region code map`** — int keys, ~215 entries, read on
every parse (`ContainsKey` up to 3× in `MaybeExtractCountryCode`, plus `TryGetValue` in
`IsValidNumber`, `GetRegionCodeForCountryCode`, `HasValidCountryCallingCode`). Behind a file-scoped
type alias so the per-target difference sits in one place.

**`perf: size the normalize stack buffer to the input`** — `stackalloc` zero-initializes, so the
fixed 256-char buffer from the first commit cleared 512 bytes on every call where the original
cleared only `number.Length * 2`.

### Build and CI

**`chore: drop the out-of-support net9.0 target`** — there are zero `NET8_0`/`NET9_0`/`NET10_0`
conditionals, so those three assets compiled from identical source. .NET 9 left support 2026-05-12;
net9 consumers resolve the net8.0 asset. Cuts ~25% off build and the test matrix.

**`chore: centralize build settings and package versions`** — `Directory.Build.props` plus Central
Package Management, collapsing dependabot's seven nuget entries to one. `LangVersion` goes
`preview` → `latest` so an SDK bump can't change the semantics of already-shipped code. Both demo
workflows' `paths:` filters gained the props files, or they'd inherit settings no PR job builds.

**`chore: keep nuget audit findings from failing the build`** — `NU1901`–`NU1904` become warnings
and `NuGetAuditMode=all` covers transitives. Otherwise a newly published advisory against the
netstandard2.0-only deps fails the *release* build days after a commit that touched nothing.

**`build: pin the sdk major and verify builds are reproducible`** — `global.json` pins .NET 10 with
`allowPrerelease: false`. New CI step packs, cleans, packs again and compares assembly hashes;
cleaning also drops the generated metadata bins, so it covers MetadataBuilder's output too.

**`ci: benchmark the base commit on the same runner as the branch`** — the PR base is checked into a
worktree and measured in the same job. The cached baseline came from a different machine, and
cross-machine variance is a few percent — the same size as the effects being measured, which is why
the previous run reported the two `ParseValidateAndFormat` sizes moving in *opposite* directions.
Removing the cache also removes the cron that existed only to keep it inside GitHub's 7-day eviction
window. ~10 min per PR instead of ~5. Missing base results now fail the job rather than posting a
one-sided comment.

**`refactor: guard nullable annotations on netstandard2.0 rather than a net version`** — six
`#if NET6_0_OR_GREATER` blocks only selected between `T?` and `T` signatures, tracking the csproj's
`<Nullable>` condition rather than any API floor. They now say `!NETSTANDARD2_0`, which is what the
csproj actually tests. The three NET6 guards that do gate real APIs (`string.Concat(ReadOnlySpan)`,
`StringBuilder.Append(StringBuilder, int, int)`, `HashSet<T>(int)`) are unchanged, as are the NET7
`[GeneratedRegex]` and NET8 Frozen guards.

### Reviewer notes

- **Benchmarks can't resolve small deltas even on one runner.** `ExtractPossibleNumber_CleanInput`
  moved ±4% between runs despite this PR not touching it — code layout and JIT effects. Read the
  parse result as credible because both input sizes agree in direction and magnitude, not because a
  single delta cleared its error bars. The `ColdStart*` benchmarks carry ±35% error and can't
  support any conclusion.
- **`GetRegionCodesForCountryCode` still returns the internal `List<string>`** as
  `IReadOnlyList<string>` — safe unless a caller casts back, where Java uses `unmodifiableList`.
  Left alone because it *is* on an internal path (`PhoneNumberOfflineGeocoder.cs:137`), so guarding
  it costs a wrapper allocation per geocode.
- **The determinism check proves same-machine reproducibility only.** It won't catch path dependence
  or filesystem enumeration order.

### Deliberately not done

- **`LocaleData.Data`** — nested `ImmutableDictionary`, two slow probes per geocode call, the largest
  read win left. Needs the change in `lib/DumpLocale.java` rather than the generated file, and
  freezing something that size is where construction cost bites. Separate PR so its numbers isolate.
- **`AnalysisMode` on the main library** (weaker than the Extensions helper today) — will produce a
  pile of CA warnings on the ported core, which `TreatWarningsAsErrors` turns into a build break.
- **Sorting the `EmbeddedResource` globs** — MSBuild may already sort glob results; confirming needs
  a cross-machine comparison.
- **Broad C# syntax modernization** in the ported core — makes future upstream Java diffs harder to
  read for cosmetic gain.


wmundev (approved):
nice changes, looks good!

PR #380, opened 2026-08-05
Reuse the parse scratch buffers per thread

twcclegg:
#378's phase split showed `Parse` holds **351 KB of the 580 KB** per 1000 numbers (60%) while being
the *fastest* phase — `Format` is 2.4× slower but only 32% of allocation. So the allocation work
belongs in `Parse`, not the regex chain I'd have guessed at.

Of that ~351 B per number, roughly half is two `StringBuilder`s: `ParseHelper` allocates one for the
national number and another for its normalized form on every call, each an object plus a chunk.

They're now taken from a small per-thread cache, so a steady stream of parses on one thread reuses
the same two buffers. No method signatures change.

Reentrancy-safe by construction: `Acquire` nulls the slot it takes from, so a nested parse — `Parse`
is reachable from `IsNumberMatch` and the example-number helpers — gets a fresh builder rather than
sharing one still in use. Buffers are returned only on the success path; throw paths leave them to
the collector, costing one allocation next call and nothing else.

Watch `ParseOnly`'s Allocated column: BenchmarkDotNet measures that deterministically, so it's the
number that settles whether this worked.

### DIFF
--- a/csharp/PhoneNumbers/PhoneNumberUtil.cs
+++ b/csharp/PhoneNumbers/PhoneNumberUtil.cs
@@ -2743,7 +2743,7 @@ namespace PhoneNumbers
             if (numberToParse.Length > MAX_INPUT_STRING_LENGTH)
                 throw new NumberParseException(ErrorType.TOO_LONG, "The string supplied was too long to parse.");
 
-            var nationalNumber = new StringBuilder();
+            var nationalNumber = StringBuilderCache.Acquire(numberToParse.Length);
             BuildNationalNumberForParsing(numberToParse, nationalNumber);
 
             var nationalNumberString = nationalNumber.ToString();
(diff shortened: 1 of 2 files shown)


twcclegg:
Measured net-negative, closing.

`ParseOnly`: allocation 350.95 -> 340.76 KB (**-10 B per number**, against ~190 B predicted), and time 434.9 -> 463.8 us (**+6.6%**, error bars around 0.7 us so well outside noise).

Small short-lived StringBuilders were already cheap — gen0 allocation is a pointer bump — so pooling them bought ~3% of the allocation while adding a thread-static lookup and capacity check to every parse.

Useful negative result: it rules out the two StringBuilders as the bulk of Parse's ~351 B per number, which was the working assumption. Recorded so it isn't retried.

PR #387, opened 2026-08-07
Close the Scorecard token-permission and unpinned-restore findings

twcclegg:
Closes the two actionable OpenSSF Scorecard findings (score is 7.3 as of today).

**Token-Permissions 0/10** — three warnings, all structural rather than actual over-grant:
- `codeql.yml` had no top-level `permissions:` block, so the file defaulted to whatever the repo
  default is. Now pinned to `contents: read`; the job's own block is unchanged.
- `create_new_release_on_new_metadata_update.yml` declared `contents: write` + `actions: write` at
  file level. Both moved to the job, so a workflow-wide write token is never minted. Same
  effective grant, same comment explaining why the writes are needed.

`post_performance_test_comment.yml` and `deploy-demo.yml` also carry top-level writes but were not
flagged — Scorecard only treats contents/actions/packages writes as high severity, and both of
those files have a single job, so moving them would change nothing. Left alone.

**Pinned-Dependencies 8/10** — "1 of 3 nugetCommand dependencies pinned". Adds `--locked-mode` to
the demo and publish restores, matching what the unit-test workflow already does. All eight
projects have a committed `packages.lock.json`, so this is enforcement, not new policy. It matters
most in `publish_nuget.yml`, which is the restore whose output ships.

**Branch-Protection −1** — the default `GITHUB_TOKEN` cannot read classic branch-protection rules.
Wires `repo_token` to a `SCORECARD_TOKEN` secret, falling back to `github.token` when the secret is
absent, so this is safe to merge before the PAT exists — the check just stays inconclusive until then.

### DIFF
--- a/.github/workflows/build_and_run_demo_tests.yml
+++ b/.github/workflows/build_and_run_demo_tests.yml
@@ -34,8 +34,10 @@ jobs:
         uses: actions/setup-dotnet@a98b56852c35b8e3190ac28c8c2271da59106c68 # v5
         with:
           dotnet-version: 10.x
+      # --locked-mode fails the restore rather than silently resolving a different transitive
+      # version than packages.lock.json records.
       - name: Restore dependencies
-        run: dotnet restore PhoneNumbers.Demo.Tests/PhoneNumbers.Demo.Tests.csproj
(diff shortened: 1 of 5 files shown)


wmundev (approved):
looks good, nice one

PR #401, opened 2026-08-24
fix: narrow TryParse/TryParseValid catch to NumberParseException

twcclegg:
## Summary
- `PhoneNumbers.Extensions.PhoneNumber.TryParse`/`TryParseValid` used a bare `catch { }`, silently swallowing any exception (not just parse failures) and masking bugs.
- Narrows both to `catch (NumberParseException)`, matching the exception type `PhoneNumberUtil.Parse` actually throws.

## Test plan
- [x] `dotnet build csharp --no-restore` — clean, 0 warnings/errors
- [x] `dotnet test csharp/PhoneNumbers.slnx` — 864 tests passed (net8.0 + net10.0), including the existing `TryParse("1235557704", "ZX", ...)` case that already exercises the narrowed exception path

### DIFF
--- a/csharp/PhoneNumbers.Extensions/PhoneNumber.cs
+++ b/csharp/PhoneNumbers.Extensions/PhoneNumber.cs
@@ -11,7 +11,7 @@
                 number = PhoneNumberUtil.Parse(input, region);
                 return true;
             }
-            catch
+            catch (NumberParseException)
             {
                 number = null;
                 return false;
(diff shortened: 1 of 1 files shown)


twcclegg:
Review confirmed this narrowing is correct and exposed (rather than caused) a pre-existing bug in `PhoneNumberUtil.Parse` — fixed in #406. Recommend merging #406 alongside this one.

twcclegg:
Superseded by #405.

PR #402, opened 2026-08-24
fix: cap AsYouTypeFormatter growth and guard PhoneNumberUtil against null

twcclegg:
## Summary
- `AsYouTypeFormatter.InputDigit` re-derives its output from the full accrued buffers on every call (including a regex match against the whole accumulated `nationalNumber`, plus a `ToString()` of the ever-growing `accruedInput` on most bail-out paths). With no cap on input length, per-call cost grows worse than quadratically with digits typed — measured 500k digits taking 30s+, extrapolating past two minutes. Real national significant numbers plus a parsed extension never approach 50 characters, so past that the formatter now freezes its output and every further call is O(1).
- Upstream Java has the identical shape (full `matches()` against the whole accumulated `nationalNumber` in `attemptToFormatAccruedDigits`, no length cap), so this isn't a porting bug — it's a defensive cap added here, in the same spirit as the existing extension-digit length limits, that changes nothing for any real phone number.
- Several `PhoneNumberUtil` entry points dereferenced a null `PhoneNumber` argument several frames deep instead of failing fast at the boundary, surfacing a bare `NullReferenceException` instead of a typed one. `IsViablePhoneNumber`, `Format`, `IsValidNumber`, `IsPossibleNumber`, `GetRegionCodeForNumber` (which `GetNumberType` calls first), and `IsNumberMatch` now throw `ArgumentNullException`.

## Test plan
- [x] `dotnet build csharp` — clean, 0 warnings/errors, all TFMs (netstandard2.0/net8.0/net10.0)
- [x] `dotnet test csharp/PhoneNumbers.slnx` — full matrix, 432/432 passing (net8.0 + net10.0)
- [x] Empirically verified the formatter fix: 2,000,000 digits fed one at a time now takes ~1ms (was hanging past 15s, extrapolated ~2 minutes) via an ad hoc stress harness against the built library
- [x] Verified normal formatting is unaffected (`(415) 666-7777`, `+44 117 496 0123`, etc.)
- [x] Verified all previously-crashing null-arg call sites now throw `ArgumentNullException` with the correct `ParamName`

### DIFF
--- a/csharp/PhoneNumbers.Test/TestAsYouTypeFormatter.cs
+++ b/csharp/PhoneNumbers.Test/TestAsYouTypeFormatter.cs
@@ -953,5 +953,42 @@ namespace PhoneNumbers.Test
             Assert.Equal("777777 9876 789", formatter.InputDigit('9'));
             Assert.Equal("777777 9876 7890", formatter.InputDigit('0'));
         }
+
+        [Fact]
+        public void TestInputDigit_KeepsEchoingWithinMaxInputStringLength()
+        {
+            // A plausible, if unusually long, real-world entry (national number, verbose extension,
+            // several pause characters) must not be silently truncated/frozen. This regression would
+            // previously trip once accrued input passed 50 characters: every InputDigit call from then
+            // on returned the same frozen snapshot no matter what was typed next. 250 is
+            // PhoneNumberUtil.MAX_INPUT_STRING_LENGTH - the same bound Parse() itself uses for "is this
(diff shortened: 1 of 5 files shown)


twcclegg:
Fixed both review findings: raised the cap from an arbitrary 50 chars to `PhoneNumberUtil.MAX_INPUT_STRING_LENGTH` (250, same bound `Parse()` itself uses) so realistic long entries no longer silently freeze mid-input, and `cappedOutput` is now properly `string?` instead of hidden behind `null!`. Added regression tests for both the not-frozen and still-capped-eventually behavior. 868 tests pass.

Issue #375, opened 2026-08-05
Tracking: improvements blocked on a major version rev


twcclegg:
The public API is frozen: the NuGet version tracks the upstream libphonenumber tag 1:1, so there is
no version number available to signal a break to consumers — it would ship inside what looks like a
routine metadata bump. The only opening would be upstream google/libphonenumber revving its own
major version (a `v10.y.z` tag), which nobody here controls or can predict.

This issue collects improvements that are worth making but need a public API change, so they aren't
re-derived every time someone goes looking. **Living list — add to it when something new turns up.**

## Public types that block internal optimisation

**`LocaleData.Data`** — `public static readonly ImmutableDictionary<string, ImmutableDictionary<string, string>>`

A field's type is part of its IL signature, so switching to `FrozenDictionary` gives consumers a
`MissingFieldException` at runtime rather than a compile error. A field also can't become lazy or a
property. Adding a frozen member *alongside* doesn't help: touching the type runs the static
initialiser, so both structures get built and held (~1.1 MB extra, 249 outer / 46,812 inner entries).

Worth knowing before anyone gets excited: measured, `Locale.GetDisplayCountry` is ~89 ns of a
~1.53 µs geocode call, and the dictionary probes are only part of that. **Low priority even when
unblocked.** `LocaleData` is arguably an implementation detail that should be `internal` — also
breaking.

**`PhoneNumberUtil.GetSupportedGlobalNetworkCallingCodes()`** returns `Dictionary<int, PhoneMetadata>.KeyCollection`

Pins `countryCodeToNonGeographicalMetadataMap` to a concrete `Dictionary`, blocking a
`FrozenDictionary` conversion. It is also why ~9 non-geographical metadata files are eagerly loaded
at construction instead of lazily. `IReadOnlySet<int>` would free both.

**`PhoneNumberUtil.GetSupportedRegions()`** returns `HashSet<string>`

Returning the live set let any caller `Remove()` a region and break validation for the rest of the
process (fixed in #370 by returning a copy, which costs an allocation per call). `IReadOnlySet<string>`
would remove the copy and allow the frozen set to be exposed directly. Same shape for
`GetSupportedCallingCodes()` (`HashSet<int>`) and `GetSupportedTypesForRegion()` (`HashSet<PhoneNumberType>`).

**`GetRegionCodesForCountryCode()`** returns the internal `List<string>` as `IReadOnlyList<string>`

[...]

PR #404, opened 2026-08-24
internal: BinaryPrimitives in ByteBuffer, volatile singleton instance

twcclegg:
## Summary
Two small, internal-only, non-breaking cleanups from the "explicitly not blocked" section of #375:

- `FlyweightMapStorage`'s internal `ByteBuffer` had already moved off `MemoryStream`/`BinaryReader`/`BinaryWriter` onto a plain `byte[]` (see `fadb2dfa`), so no disposable fields remain and it was never `IDisposable`. This swaps its manual little-endian bit-shifting for `System.Buffers.Binary.BinaryPrimitives.{Read,Write}{Int16,Int32}LittleEndian`, matching #375's suggestion of "rewritten over `BinaryPrimitives` on a plain `byte[]`" — same behavior, same endianness, no public/internal signature changes.
- `PhoneNumberUtil.instance` is now `volatile`, matching the double-checked lock in `GetInstance()`. The `??=` inside the lock is untouched — #375 notes `CA1508` flags it as dead code, but that's a documented false positive; removing it would reintroduce the race.

## Test plan
- [x] `dotnet build csharp --no-restore` — 0 warnings, 0 errors
- [x] `dotnet test csharp/PhoneNumbers.slnx` — 864 tests passed (net8.0 + net10.0)
- [x] `PhoneNumbers.PerformanceTest` geocoder benchmark run locally as a sanity check (no crash, no obvious regression); CI's `run_performance_tests.yml` will do the real base-vs-PR comparison

### DIFF
--- a/csharp/Directory.Packages.props
+++ b/csharp/Directory.Packages.props
@@ -18,6 +18,7 @@
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
     <PackageVersion Include="SharpFuzz" Version="2.3.0" />
     <PackageVersion Include="System.Collections.Immutable" Version="8.0.0" />
+    <PackageVersion Include="System.Memory" Version="4.6.3" />
     <PackageVersion Include="System.Text.Json" Version="8.0.6" />
     <PackageVersion Include="xunit" Version="2.9.3" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.8.2" />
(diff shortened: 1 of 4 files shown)


twcclegg:
Fixed the netstandard2.0 build break the review found: `ByteBuffer`'s `BinaryPrimitives`/`Span` usage needs `System.Memory` on that TFM. Added it as a netstandard2.0-only `PackageReference` (same pattern this project already uses for `System.Collections.Immutable`) rather than \#if-splitting the implementation, since the behavior is identical on every TFM. Verified: netstandard2.0/net8.0/net10.0 all build clean, full test suite passes (864 tests), and `dotnet pack` passes API-compat validation.

PR #405, opened 2026-08-24
feat(extensions): TryParse overloads, TypeConverter, and a real phone-number ValidationAttribute

twcclegg:
## Summary
- `PhoneNumber.TryParse`/`TryParseValid` gain a 2-arg overload (region defaults to `null`), matching the conventional .NET `TryParse(string, out T)` shape.
- `PhoneNumberTypeConverter` converts `PhoneNumber` to/from its E.164 string — for consumers that register `TypeConverter`s (config binding, `PropertyGrid`, MVC model binding, etc.). Not applied automatically; register it explicitly, same pattern as the existing `PhoneNumberConverter` for `System.Text.Json`.
- `PhoneNumberAttribute` is a `DataAnnotations` `ValidationAttribute` backed by `PhoneNumberUtil.IsValidNumber`, unlike the framework's built-in `[Phone]` which is a loose regex. Accepts a `string` (parsed via the optional `Region` property) or an already-parsed `PhoneNumber`.

Note on scope: an earlier version of this considered a `TryParse(string, IFormatProvider, out PhoneNumber)` shape aimed at ASP.NET Core minimal-API automatic parameter binding, but that only works if the method lives on `PhoneNumbers.PhoneNumber` itself (the framework reflects on the parameter's own type) — decided to keep this Extensions-only rather than touch the core faithful-port model class, so this PR is a plain convenience overload rather than automatic binding support.

`netstandard2.0` needs `System.ComponentModel.Annotations` for `ValidationAttribute`; `net8.0`/`net10.0` already have it in the shared framework — added centrally in `Directory.Packages.props`, referenced only on the `netstandard2.0` leg like the existing `System.Text.Json` reference.

## Test plan
- [x] `dotnet build csharp --no-restore` — 0 warnings/errors
- [x] `dotnet test csharp/PhoneNumbers.slnx` — 888 tests passed (net8.0 + net10.0), including new tests for all three additions
- [x] `dotnet pack -c Release csharp/PhoneNumbers.Extensions -p:VersionPrefix=9.0.38` — packs and passes `EnablePackageValidation` against the 9.0.37 baseline (confirms the additions are purely additive, no API-compat breaks)

### DIFF
--- a/csharp/Directory.Packages.props
+++ b/csharp/Directory.Packages.props
@@ -18,6 +18,7 @@
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
     <PackageVersion Include="SharpFuzz" Version="2.3.0" />
     <PackageVersion Include="System.Collections.Immutable" Version="8.0.0" />
+    <PackageVersion Include="System.ComponentModel.Annotations" Version="5.0.0" />
     <PackageVersion Include="System.Text.Json" Version="8.0.6" />
     <PackageVersion Include="xunit" Version="2.9.3" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.8.2" />
--- a/csharp/PhoneNumbers.Extensions.Test/TestExtensions.cs
+++ b/csharp/PhoneNumbers.Extensions.Test/TestExtensions.cs
@@ -45,6 +45,14 @@ namespace PhoneNumbers.Extensions.Test
         }
 
         [Fact]
+        public void TestE164_TwoArgOverload()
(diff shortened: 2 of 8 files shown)


twcclegg:
Updated after code review (8-angle pass): fixed two findings on `PhoneNumberTypeConverter`/`PhoneNumberAttribute` (see latest commit). Also worth flagging: review surfaced a real, pre-existing bug in the core parser reachable through `PhoneNumberAttribute` on malformed RFC3966 input — fixed separately in #406, which should merge alongside this one and #401.

PR #407, opened 2026-08-24
ci: exclude generated obj/ from CodeQL, document the second empty catch

twcclegg:
## Summary
- CodeQL's manual build-mode analysis was picking up the .NET regex source generator's own output under `obj/` (gitignored, rebuilt fresh every run) — not repo code, so nobody can act on `cs/unused-label` / `cs/useless-assignment-to-local` findings against Microsoft's generated `RegexGenerator.g.cs`. Added `.github/codeql/codeql-config.yml` with `paths-ignore: ["**/obj/**"]`, wired in via `config-file:` on the `init` step. This closes out ~20 open code-scanning alerts that were pure noise.
- `cs/empty-catch-block` (alert #417) flagged the second `catch (NumberParseException)` in `TestPhoneNumberProperties.ParseFailsOnlyWithNumberParseException` — it's the same intentional "anything else is a bug" pattern as the catch block directly above it (and the one in `PhoneNumbers.Fuzz/Program.cs`), just missing that block's explanatory comment. Added it.

Found via `gh api repos/.../code-scanning/alerts` while auditing the repo's existing tool findings — both are config/comment-only, no behavior change.

## Test plan
- [x] `dotnet build csharp` — clean, 0 warnings/errors, all TFMs
- [x] `dotnet test csharp/PhoneNumbers.slnx` — full matrix, 432/432 passing (net8.0 + net10.0)
- [x] Validated both YAML files parse correctly

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,8 +1,22 @@
 name: "CodeQL config"
 
-# obj/ is gitignored and rebuilt fresh by the manual `dotnet build` step in codeql.yml - it isn't
-# repo code, just the .NET regex source generator's own output (RegexGenerator.g.cs) picked up
-# by build-mode: manual. Without this, CodeQL reports cs/unused-label and
-# cs/useless-assignment-to-local against Microsoft's generated code, which nobody here can act on.
-paths-ignore:
-  - "**/obj/**"
+# paths-ignore doesn't work here: for build-mode: manual on a compiled language, CodeQL analyzes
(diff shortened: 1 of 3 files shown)


twcclegg:
The paths-ignore approach doesn't work for build-mode: manual (confirmed against GitHub's own docs) — switched to query-filters disabling cs/unused-label and cs/useless-assignment-to-local outright, the only mechanism that actually suppresses these for a compiled-language manual build. Real tradeoff: those two rules stop firing repo-wide, not just on generated code — accepted since both are low-severity style checks.

PR #410, opened 2026-08-24
test: exclude the ported protobuf-style Builder classes from coverage

twcclegg:
## Summary
`PhoneMetadata/Builder`, `PhoneNumber/Builder`, `NumberFormat/Builder`, `PhoneNumberDesc/Builder`, and `PhoneMetadataCollection/Builder` are dozens of near-identical Get/Set/Has/Clear blocks per field (1400+ lines for `PhoneMetadata/Builder` alone), so their sheer size swamped the project coverage percentage without the per-field wiring itself carrying real verification value.

Their one piece of actual behavior — `MergeFrom`/`Build`/`BuildPartial` — is already exercised by `TestBuildMetadataFromXml.cs`, the only place these builders run outside the build-time metadata pipeline (the default load path reads pre-built binaries, not XML, so the rest of the suite never touches them). Excluding them locally: overall line-rate went from 0.8144 to 0.8856, confirming they were the drag rather than a real gap.

Verified empirically against coverlet's actual filter matching rather than assumed — `[Module]Type.Builder` and `[Module]Type/Builder` both silently matched nothing; `[PhoneNumbers]*Builder` is what actually excludes the five nested types (confirmed via a before/after class-list diff) without touching the outer message classes.

## Test plan
- [x] `dotnet build csharp` — clean, 0 warnings/errors, all TFMs
- [x] `dotnet test csharp/PhoneNumbers.slnx -p:TargetFrameworks=net10.0` — 432/432 passing
- [x] Verified via a before/after cobertura class-list diff that exactly the 5 intended `*Builder` types are excluded and nothing else (`PhoneMetadata`, `PhoneNumberUtil`, etc. unaffected)

### DIFF
--- a/csharp/coverlet.runsettings
+++ b/csharp/coverlet.runsettings
@@ -5,18 +5,27 @@
       <DataCollector friendlyName="XPlat code coverage">
         <Configuration>
           <!-- CountryCodeToRegionCodeMap: generated data table, not code under test.
-               *Builder (matches the five nested PhoneMetadata/Builder, PhoneNumber/Builder,
-               NumberFormat/Builder, PhoneNumberDesc/Builder, PhoneMetadataCollection/Builder
-               types): hand-ported protobuf-style builders. Each is dozens of near-identical
-               Get/Set/Has/Clear blocks per field (over a thousand lines for
-               PhoneMetadata/Builder alone), so their line count swamps the project percentage
-               without the per-field wiring itself carrying verification value - nobody is going
-               to hand-test that a setter's value round-trips through its getter. The behavioral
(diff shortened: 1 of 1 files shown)


twcclegg:
Fixed: scoped the exclusion to the five intended types (`Outer*Builder` per type) instead of a blanket `*Builder`. Note for future reference: coverlet's XPlat collector doesn't match an exact `Outer+Builder`/`Outer/Builder` qualified nested-type name here — only the wildcard form works, verified by actually collecting coverage and inspecting coverage.cobertura.xml (a wrong filter fails silently, no build/test error). Confirmed identical totals to the original wildcard (7151/8078 lines, net10.0).

Issue #313, opened 2026-01-19
SA mobile numbers starting with 579 are rejected by IsValidNumber (IsPossibleNumber=true)


abdulatef2009:
## Summary
In Saudi Arabia (region "SA"), some mobile numbers starting with `579` are parsed successfully but are marked as invalid by `IsValidNumber()`, while similar numbers like `569...` are valid.

This causes false negatives in production, so we temporarily hotfixed using `IsPossibleNumber()` as a fallback.

## Steps to reproduce
Using libphonenumber-csharp on .NET 8:

```csharp
using PhoneNumbers;

var util = PhoneNumberUtil.GetInstance();

// Example numbers (same length/pattern, only prefix differs)
var n1 = util.Parse("579293970", "SA");
var n2 = util.Parse("569293970", "SA");

Console.WriteLine($"579... IsPossible={util.IsPossibleNumber(n1)}, IsValid={util.IsValidNumber(n1)}, Type={util.GetNumberType(n1)}");
Console.WriteLine($"569... IsPossible={util.IsPossibleNumber(n2)}, IsValid={util.IsValidNumber(n2)}, Type={util.GetNumberType(n2)}");

twcclegg:
That behavior matches Google's behavior here
https://libphonenumber.appspot.com/phonenumberparser?number=579293970&country=SA
https://libphonenumber.appspot.com/phonenumberparser?number=569293970&country=SA
You'll need to take the issue up with them. Please follow the instructions here https://github.com/twcclegg/libphonenumber-csharp/blob/main/.github/ISSUE_TEMPLATE/bug_report.md

Issue #272, opened 2024-11-09
New phone number format for Benin not supported


bertsch-ronja-office:
Benin changes its phone number format - beginning 30 November 2024, all existing telephone numbers in Benin will add "01" in front of them.

Source: https://en.wikipedia.org/wiki/Telephone_numbers_in_Benin

The newest version of libphonenumber-csharp (8.13.49) does rate the new format (example phone number +2290140123456) as invalid (whereas the old format +22940123456) works.

wmundev:
hi @bertsch-ronja-office , try the new version, it should work now - 
http://libphonenumber.appspot.com/phonenumberparser?number=%2B2290140123456

https://github.com/twcclegg/libphonenumber-csharp/releases/tag/v8.13.51

wmundev:
closed as resolved with the new metadata release, feel free to reopen if its still not resolved

bertsch-ronja-office:
@wmundev  Thanks, the new format works now.

Still we are having an issue with parsing the E164 format - in the link you sent me the E164 is: +2290140123456


But using the IsE164 method, the leading 0 is stripped, it returns the following as E164: +229140123456

var e164PhoneNumber = phoneNumberUtil.Format(phoneNumber, PhoneNumberFormat.E164);

What is now the right E164 number?

Issue #259, opened 2024-07-22
New regulation for phone number prefixes in Australia


leonardonl:
There is a new Australian regulation concerning phone number prefixes:
Phone numbers starting with 04{0-8} are accepted, but 049 and 05{1-9} are rejected, and according to the [AUS telecom numbering plan](https://www.legislation.gov.au/F2015L00319/latest/text), 04{0-9} and 05{0-9} must be accepted. A similar situation happens when you try +6149 or +615{0-9} (or +61049 and +6105{0-9}).

twcclegg:
Metadata for this library is controlled by Google. See https://github.com/twcclegg/libphonenumber-csharp/blob/main/.github/ISSUE_TEMPLATE/bug_report.md for opening an issue with them.

Issue #214, opened 2023-12-15
French mobile numbers with prefix "07" are not detected as mobile numbers


madmox:
French mobile numbers can start with "06" or "07" ([source](https://en.wikipedia.org/wiki/Telephone_numbers_in_France)). The library only detects mobile numbers with prefix "06".

Example:
```csharp
using PhoneNumbers;

PhoneNumberUtil PARSER = PhoneNumberUtil.GetInstance();

PhoneNumber parsedNumber1 = PARSER.Parse("+33612345678", defaultRegion: null);
Console.WriteLine(PARSER.GetNumberType(parsedNumber1)); // => MOBILE

PhoneNumber parsedNumber2 = PARSER.Parse("+33712345678", defaultRegion: null);
Console.WriteLine(PARSER.GetNumberType(parsedNumber2)); // => UNKNOWN
```

`libphonenumber-csharp` version: `8.13.26`

twcclegg:
This behavior seems to match https://libphonenumber.appspot.com/phonenumberparser?number=%2B33712345678, as such you'll need file an issue with Google https://groups.google.com/g/libphonenumber-discuss

madmox:
[Issue](https://issuetracker.google.com/issues/316476490) created.

PR #422, opened 2026-08-26
Add automated triage bot for metadata-only issues

twcclegg:
## Changes
- Add `.github/workflows/triage_metadata_issues.yml`, which runs on every newly opened issue and uses the GitHub Copilot CLI (`actions/ai-inference` with the `copilot` provider) to classify whether it's actually a report about upstream Google libphonenumber *metadata* — an unrecognized prefix, an outdated numbering plan, a new country code — rather than a bug in this port's C# code.
- Looking through the issue history, this pattern shows up constantly (#313, #272, #259, #214, #173, #153, #143, #118, #112, #104, ...) and maintainers have been closing these by hand for years with essentially the same reply pointing to [google/libphonenumber](https://github.com/google/libphonenumber). The classifier's prompt is grounded in that history (both the closed "yes, file it upstream" examples and the closed "no, this was actually a code bug" examples like #183/#206/#212).
- When the classifier is confident it's a metadata issue, the workflow labels it `metadata` (creating the label if needed), posts a canned comment explaining why and linking to the upstream repo + the `bug_report.md` checklist, and closes the issue (`state_reason: not_planned`). The comment explicitly says the flag was automatic and to reopen if it's a false positive.
- Skips issues opened by the maintainers themselves (`OWNER`/`MEMBER`/`COLLABORATOR`), and fails open: if the CLI errors or its response can't be parsed, or the confidence comes back `low`, the issue is left untouched for a human.
- Documented the new workflow in the README under a new "Automated triage of metadata issues" section.

### Setup required after merge
This uses the modern (PAT-free) way of running Copilot CLI in Actions: `permissions: copilot-requests: write` plus the built-in `GITHUB_TOKEN`. That requires GitHub Copilot CLI usage to be enabled for the `twcclegg` account (the "Allow use of Copilot CLI" policy) — see https://docs.github.com/en/copilot/how-tos/copilot-cli/use-copilot-cli-in-actions. No repository secret needs to be created.

## Test plan
- [ ] Validated the workflow YAML parses and the embedded `github-script` blocks are syntactically valid JavaScript.
- [ ] Confirmed `actions/ai-inference` pinned SHA corresponds to its `v3` tag (the current Copilot-CLI-only major version — `v1`/`v2` default to the now-retired GitHub Models API and would not work).
- [ ] Can't fully exercise the `issues: opened` trigger from a PR — once merged, open a throwaway test issue that mirrors e.g. #313 ("SA mobile numbers rejected") to confirm it gets classified, labeled, commented, and closed automatically, and a normal bug/question issue to confirm it's left alone.

---


twcclegg:
Pushed a follow-up commit per feedback:

- Replaced the handful of hardcoded examples in the classifier's system prompt with two exhaustive, human-verified files: [`metadata_examples.md`](https://github.com/twcclegg/libphonenumber-csharp/blob/claude/metadata-issue-bot-1bkszq/.github/triage/metadata_examples.md) (30 confirmed "redirect to Google" issues) and [`not_metadata_examples.md`](https://github.com/twcclegg/libphonenumber-csharp/blob/claude/metadata-issue-bot-1bkszq/.github/triage/not_metadata_examples.md) (26 confirmed real bugs/questions/build issues), both compiled by reading through this repo's closed-issue comment history rather than guessing.
- The workflow now closes each list into the system prompt at runtime (`cat` into a temp file, fed via `system-prompt-file`).
- Added a step that, whenever the bot closes a new issue as a metadata issue, opens a PR appending that issue to `metadata_examples.md` — so the classifier's grounding keeps growing from real outcomes instead of staying frozen at this snapshot. That needed `contents: write` and `pull-requests: write` added to the workflow's permissions.

---
_Generated by [Claude Code](https://claude.ai/code)_

twcclegg:
Pushed another update per feedback:

- Dropped `not_metadata_examples.md` entirely.
- `metadata_examples.md` is now the single source of truth, framed explicitly: every issue #1–#422 is either a PR (issues and PRs share one number sequence in this repo), a listed metadata issue, or — by not appearing in the list — an implied non-metadata issue. No third bucket.
- Went back through the rest of the closed-issue history to make the list genuinely exhaustive rather than a curated sample — caught several I'd missed the first pass (#14, #25, #109, #149, #165, #181, #91), each verified against its actual resolution comment. 30 → 39 entries.
- Updated `system_prompt.md` and the workflow's `cat` step to match (only two files get concatenated into the classifier prompt now).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #426, opened 2026-08-26
fix: land CodeQL fixes orphaned by a merge race on PRs #412-#418

twcclegg:
## Summary
PRs #412, #413, #414, #415, #416, #418 got merged while I was still pushing CodeQL cleanup commits to those same branches in a follow-up session, so each branch ended up exactly one commit ahead of the already-merged `main` — the fix commits never made it in. This PR cherry-picks all six orphaned commits (verified: no file overlap between them, no conflicts on cherry-pick):

- `style: map region codes via Select in GetExpectedCost` (ShortNumberInfo.cs, ex-#412)
- `style: filter local-only lengths explicitly via Where` (BuildMetadataFromXml.cs, ex-#413)
- `style: drop redundant int.GetHashCode() call in PhoneNumberMatch` (PhoneNumberMatch.cs, ex-#414)
- `style: ternary builder merges and drop redundant int.GetHashCode() calls` (Phonemetadata.cs, PhoneNumberDesc.cs, ex-#415)
- `style: drop always-true lower bound in IsLatinLetter's BASIC_LATIN check` (PhoneNumberMatcher.cs, ex-#416)
- `style: address remaining CodeQL findings in PhoneNumberUtil.cs` (PhoneNumberUtil.cs, ex-#418)

Each was originally reviewed against the CodeQL alert it addresses (and against upstream Java where relevant) in the earlier PRs; this is purely landing what was already vetted, not new work.

## Test plan
- [x] `dotnet build csharp/PhoneNumbers/PhoneNumbers.csproj -c Release` — all three TFMs (netstandard2.0/net8.0/net10.0) build clean
- [x] `dotnet test csharp/PhoneNumbers.slnx -p:TargetFrameworks=net10.0` — 450/450 tests pass
- [x] `git diff origin/main --stat` confirms the diff is exactly these 6 commits' changes, nothing else

### DIFF
--- a/csharp/PhoneNumbers/BuildMetadataFromXml.cs
+++ b/csharp/PhoneNumbers/BuildMetadataFromXml.cs
@@ -623,14 +623,13 @@ namespace PhoneNumbers
             // We check that the local-only length isn't also a normal possible length (only relevant for
             // the general-desc, since within elements such as fixed-line we would throw an exception if we
             // saw this) before adding it to the collection of possible local-only lengths.
-            foreach (var length in localOnlyLengths)
-                if (!lengths.Contains(length))
-                    if (parentDesc is null || parentDesc.possibleLengthLocalOnly_.Contains(length)
-                        || parentDesc.possibleLength_.Contains(length))
-                        desc.possibleLengthLocalOnly_.Add(length);
-                    else
-                        throw new Exception(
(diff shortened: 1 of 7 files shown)


twcclegg:
Superseded — these 6 commits now live on PR #423 instead (cherry-picked cleanly, verified byte-identical to what was here, full suite green). Closing without merging.

PR #446, opened 2026-08-30
perf: narrow \d to ASCII [0-9] in metadata-matched regex patterns

twcclegg:
## Changes
- Narrows `\d` to `[0-9]` in every metadata-derived regex pattern that's actually matched against phone-number input: territory `leadingDigits`, `internationalPrefix`, `nationalPrefixForParsing`, `numberFormat`'s `pattern` attribute (national and the `intlNumberFormat_` copy), per-format `leadingDigits`, and `nationalNumberPattern`.
- **Why this is safe**: .NET's `\d` (without `RegexOptions.ECMAScript`) matches the full Unicode "decimal digit" category — hundreds of non-ASCII characters — which is both broader and more expensive to match than a plain `[0-9]` range check. Upstream Java's `\d` defaults to ASCII-only (`[0-9]`) unless `Pattern.UNICODE_CHARACTER_CLASS` is set, and these patterns were authored/verified against that ASCII-only default, so this also closes a semantic gap from upstream, not just a perf one.
- **Why it's correct**: every one of these patterns is only ever matched against input that's already passed through `PhoneNumberUtil.Normalize`/`NormalizeDigits` (which converts fullwidth/other-script digits to ASCII) or against `GetNationalSignificantNumber`, which is rebuilt from the numeric `NationalNumber` proto field and is ASCII by construction. Traced through the parse path (`MaybeStripInternationalPrefixAndNormalize`, `ParseHelper`/`MaybeStripNationalPrefixAndCarrierCode`), `AsYouTypeFormatter` (`NormalizeAndAccrueDigitsAndPlusSign`), and `PhoneNumberMatcher` (`ParseAndVerify`/`ParseAndKeepRawInput`) — full detail in commit `a1204a2`. `PhoneNumberMatcher`'s free-text candidate-scanning regexes and `PhoneNumberUtil`'s extension-parsing pattern are hand-written, not XML-metadata-derived, and are deliberately left untouched since they run on raw, un-normalized text.
- The rewrite (`BuildMetadataFromXml.NarrowDigitClassToAscii`) is structure-aware, not a blind string replace: it tracks escape pairs and character-class entry/exit so `\d` already nested inside a class (e.g. `[\d-]`) narrows to `[0-9-]` rather than producing broken nested brackets, and handles the POSIX literal-`]`-as-first-member idiom. `\D` narrows to `[^0-9]` outside a class; inside a class it throws rather than silently mishandling it (doesn't occur in shipped metadata today — verified by grep across all four resource XML files).

## Test plan
- [x] `dotnet build csharp` — clean on all three TFMs (netstandard2.0, net8.0, net10.0), zero warnings under `TreatWarningsAsErrors` + trim/AOT analyzers
- [x] `dotnet test csharp/PhoneNumbers.slnx` — 434 + 37 tests pass on net8.0 and net10.0, including new `TestNarrowDigitClassToAscii` theory tests covering real patterns from `resources/PhoneNumberMetadata.xml`, the nested-class case, `[^\d]`, POSIX-literal-bracket shapes, escaped-backslash-before-`d`, and a match-semantics-preservation check
- [x] `dotnet run -c Release --framework net10.0 -- --filter "*PhoneNumberWorkflowBenchmark*"` / `*ColdStartBenchmark*` — modest, directionally consistent steady-state win (~5-9% on `ParseOnly`/`ParseNationalFormat`/`ValidateOnly`/`FormatOnly`, several exceeding their reported confidence margins), cold start flat as expected (this is a matching-cost change, not a JIT-cost change)

---

### DIFF
--- a/csharp/PhoneNumbers.Test/TestBuildMetadataFromXml.cs
+++ b/csharp/PhoneNumbers.Test/TestBuildMetadataFromXml.cs
@@ -78,6 +78,86 @@ namespace PhoneNumbers.Test
             Assert.Equal(validPattern, BuildMetadataFromXml.ValidateRE(validPattern, false));
         }
 
+        // Tests NarrowDigitClassToAscii(): the character-class-aware \d -> [0-9] rewrite used for
+        // metadata patterns that are actually matched against (already ASCII-normalized) input --
+        // see ValidateAndNarrowPatternRE's doc comment for exactly which fields that is.
+        [Theory]
+        // Real nationalNumberPattern values, pulled verbatim from resources/PhoneNumberMetadata.xml.
+        [InlineData("6\\d{4}", "6[0-9]{4}")]
+        [InlineData("[2-47]\\d{4}", "[2-47][0-9]{4}")]
+        // Real internationalPrefix values.
+        [InlineData("(?:0|1(?:1[0-69]|2[02-5]|5[13-58]|69|7[0167]|8[018]))0",
+            "(?:0|1(?:1[0-69]|2[02-5]|5[13-58]|69|7[0167]|8[018]))0")]
+        [InlineData("0(?:0|1[3-9]\\d)", "0(?:0|1[3-9][0-9])")]
+        // Real nationalPrefixForParsing values (note the trailing "$" anchor and alternation).
+        [InlineData("(000[2569]\\d{4,6})$|(?:(?:003768)0?)|0",
+            "(000[2569][0-9]{4,6})$|(?:(?:003768)0?)|0")]
+        [InlineData("(5\\d{6})$|1", "(5[0-9]{6})$|1")]
+        // Real numberFormat "pattern" attribute values.
+        [InlineData("(\\d)(\\d{2,3})(\\d{2})(\\d{2})", "([0-9])([0-9]{2,3})([0-9]{2})([0-9]{2})")]
+        [InlineData("(\\d)(\\d{2})(\\d{3,4})", "([0-9])([0-9]{2})([0-9]{3,4})")]
+        // \d already inside a character class must substitute the bare "0-9" in place, not wrap a
+        // second, nested set of brackets around it.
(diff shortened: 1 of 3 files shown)


twcclegg:
## Local benchmark numbers

The automated performance comment only posts when `compare-benchmarks.js`'s Welch's t-test clears a 20% relative-delta floor (see `PhoneNumbers.PerformanceTest/README.md`), and this change doesn't hit that — it's a real but modest steady-state win, not a dramatic one. Posting the local numbers manually since CI's own comment won't show them.

`PhoneNumberWorkflowBenchmark` (net10.0, Release, n=1000, single run before/after — not the multi-run statistical comparison CI does, so treat these as directional):

| Method | Before | After | Δ |
|---|---:|---:|---:|
| ParseValidateAndFormatPhoneNumbers | 3656.6 us | 3641.1 us | -0.4% (noise) |
| ParseOnly | 615.2 us | 579.6 us | **-5.8%** |
| ParseNationalFormat | 1238.1 us | 1127.3 us | **-8.9%** |
| ParseWithExtension | 1633.9 us | 1676.8 us | +2.6% (noise — extension parsing uses the untouched hand-written `ExtnPattern`, so there's no mechanism for this one to improve) |
| ValidateOnly | 989.0 us | 941.4 us | **-4.8%** |
| FormatOnly | 1530.0 us | 1438.2 us | **-6.0%** |

`ColdStartBenchmark` — flat, as expected (this narrows a character class matched during steady-state parsing/validation/formatting, not anything on the metadata-load or first-touch path):

| Case | Before | After |
|---|---:|---:|
| CreateInstance | 613.4 us ± 209.8 | 529.9 us ± 153.2 (within noise) |
| CreateInstanceAndLoadAllRegions | 11653.0 us | 11649.0 us |
| FirstRegionLookup | 613.5 us | 616.2 us |

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #453, opened 2026-09-04
feat: add DocFX API reference site, published alongside the demo

twcclegg:
## Summary

A generated API reference site for both packable projects, built with [DocFX](https://dotnet.github.io/docfx/) from the XML doc comments they already emit, and themed with the demo's own design tokens so the two sites read as one product rather than two tools bolted together.

- **Published** into the same GitHub Pages deployment as the demo, under `/docs/` — one deployment, one domain, no separate hosting.
- **Cross-linked**: the demo's sidebar gets an "API Docs" entry; the docs home page links back to the demo, GitHub and NuGet, and carries the install command with a copy-to-clipboard button.
- **Articles**: `docs/api-differences-from-java.md` becomes a page rather than a link buried in the README. `docfx/build.sh` copies it in and rewrites its repo-relative links to GitHub permalinks, since the published site ships only rendered articles.
- **Curated surface**: `filterConfig.yml` drops compiler-generated and `[EditorBrowsable(Never)]` members. Twelve public-but-internal types — metadata plumbing and prefix-map internals, unreachable from any consumer entry point — are now marked `[EditorBrowsable(Never)]` too, taking the reference from 43 documented types to 34. Package validation confirms that is not an API-compat break.

Build it with `./docfx/build.sh`, never `docfx` directly: the script restores what docfx's MSBuild pass needs and copies the articles in, so a local run, the preview workflow and the deploy all start from the same state.

One deliberate coupling worth a reviewer's attention: the docs build in `deploy-demo.yml` is **not** `continue-on-error`. A Pages deployment replaces the whole site, so carrying on past a docs failure would publish a `wwwroot` with no `docs/` in it and take `/docs` down to a 404 rather than leaving it stale. Failing the job keeps the previous deployment serving, demo and docs together — at the cost of a docfx regression also blocking demo-only deploys.

## Reviewing it

`docs_preview.yml` uploads the rendered site as an artifact on every relevant PR — no merge or Pages publish needed. Download `api-docs-preview` from the run summary, unzip, and serve it with any static server (`python3 -m http.server` from inside the folder). Opening `index.html` over `file://` won't work; the site fetches its nav and search data over HTTP.

## Test plan

- [ ] Eyeball the preview artifact in light and dark mode against the [demo](https://twcclegg.github.io/libphonenumber-csharp/)
- [ ] On merge, confirm `deploy-demo.yml` succeeds and `/docs/` is live and themed


docfx/docfx.json
```diff
@@ -0,0 +1,61 @@
...
+        "files": [ "*.md", "toc.yml" ]
+      },
+      {
+        "files": [ "articles/**.md", "articles/toc.yml" ]
+      }
+    ],
+    "resource": [
+      {
+        "//": "Shared with the Blazor demo, which publishes the same three files into its own wwwroot; kept at the repo root so the two sites cannot drift. Mapped to images/ so _appLogoPath and _appFaviconPath below stay site-relative.",
+        "src": "../assets/brand",
```
wmundev: not sure about this but i think the demo doesn't references these assets/brand files at all

docfx/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+<div class="docs-hero">
+
+<span class="docs-hero__badge"><span class="docs-hero__dot"></span> API Reference</span>
+
+# libphonenumber-csharp
+
+C# port of Google's [libphonenumber](https://github.com/google/libphonenumber) &mdash; parse, format, validate, and geolocate international phone numbers.
+
+<a class="docs-btn" href="xref:PhoneNumbers">
```
wmundev:
```suggestion
<a class="docs-btn" href="api/PhoneNumbers.html">
```

you can do this so it renders as a button, not as normal link

Image

wmundev (approved):
@twcclegg looks very good, i've tested it locally too and the site renders fine

docfx/docfx.json
twcclegg: Good catch — you're right, and the comment was just wrong. The demo has no favicon link at all and draws its rail logo from `Components/Icons/`, so docfx was the only thing reading those files.

Moved them to `docfx/brand/`, corrected the comment, and dropped the `assets/brand/**` path filters from `deploy-demo.yml` and `docs_preview.yml` since `docfx/**` now covers them. No more repo-root `assets/` directory.

docfx/index.md
twcclegg: Thanks — confirmed the cause: docfx replaces an `<a>` whose href is an `xref:` with its own generated anchor, which drops the `docs-btn` class and the arrow SVG.

Took your fix with one tweak: pointed it at `api/PhoneNumbers.yml` rather than `api/PhoneNumbers.html`. Both render the button correctly, but the literal `.html` trips an `InvalidFileLink` warning since docfx validates file links against source files — the `.yml` is the real source and docfx rewrites it to `.html` on build. Clean build is back to 0 warnings.

PR #455, opened 2026-09-05
fix: demo initial-load focus ring, theme-synced favicon, and icon/alignment polish

twcclegg:
Four changes to the demo site. The focus fix is the substantial one; the other three are small visual corrections found while checking it in a browser. Happy to split them out if you'd rather review them separately.

## 1. Don't focus the `h1` on initial page load

On first load of the demo site, the hero heading **"Phone numbers, handled."** is drawn with an orange focus ring around it. Clicking anywhere dismisses it, and it doesn't come back until the page is reloaded.

`App.razor` used Blazor's `<FocusOnNavigate RouteData="@routeData" Selector="h1" />`. Its JS side does `element.tabIndex = -1; element.focus()` on the first matching `h1` after **every** route render — including the initial page load. The element then matches `:focus-visible`, and the global rule in `wwwroot/css/app.css` paints `outline: var(--ring-w) solid var(--accent)` around it. The first click clears the browser's focus-visible flag, which is why it never returns without a reload.

Verified against the unfixed build in headless Chrome:

```
INITIAL LOAD: activeElement=H1.hero__title, tabindex="-1",
              :focus-visible=true, outline="solid 3px rgb(251,146,60)"
```

`FocusOnRouteChange` wraps `FocusOnNavigate` and renders it only once a real page navigation has happened:

- **Initial load** — no focus is moved. The browser already starts the user at the top of the document, so there is nothing to gain and a stray focus ring to lose.
- **SPA navigation to a different page** — unchanged. Focus still moves to the new page's `h1`, which is the accessibility behaviour the component exists for, so keyboard and screen reader users still land in the new content.

It compares `RouteData.PageType` across renders rather than the URI. Every page in this demo syncs its input state into the query string on blur/change (`SyncUrl` in `Home.razor`, `Formatting.razor`, `GeoTimezone.razor`, `ParseValidate.razor`, `FindNumbers.razor`, and `LiveFormatter.razor`), which calls `NavigateTo` for the *same* page — a URI comparison treated that as the first "navigation," mounting `FocusOnNavigate` on ordinary form interaction and stealing focus back to the heading, away from the field the user had just used. `PageType` only changes on an actual different-page transition, which is the same signal `FocusOnNavigate` uses internally, so this mirrors its own logic instead of guessing at it from the URI.

Same check on the fixed build:

```
INITIAL LOAD:   activeElement=BODY, tabindex=null, outline="none"
AFTER SPA NAV:  activeElement=H1.page-header__title, outline="solid 3px rgb(251,146,60)"
```

## 2. A real favicon, kept in sync with the theme toggle

The site shipped without one, so browsers fell back to the default document icon. Adds the phone logomark on the same gradient badge the sidebar uses, in three files under `assets/brand/`:

- `favicon.svg` carries a `prefers-color-scheme` media query so first paint is right before any script runs.
- `logo.svg` / `logo-dark.svg` are the resolved light and dark variants. `phoneDemo.setTheme` swaps between them, because the page's own theme toggle is something `prefers-color-scheme` can't see.

They live at the repo root rather than in `wwwroot/` because #453 publishes the same three files as the docfx site's logo and favicon. One copy means the two sites can't drift, and — since both branches add byte-identical files at the same path — whichever merges second combines cleanly rather than conflicting. A `Content`/`Link` item in `PhoneNumbers.Demo.csproj` puts them where the static web assets SDK looks, so they still ship in `wwwroot/`.

## 3. NuGet sidebar icon

The "NuGet" link used a generic stroked box outline. Replaced with NuGet's actual logomark, plus a `sidebar__link-icon--filled` modifier, since brand marks are filled shapes and the surrounding icons are stroke icons.

## 4. Install-block prompt alignment

`.install-block` used `align-items: center`, which centred the `$` against the full height of the block. On a command long enough to wrap, the prompt floated to the vertical middle instead of sitting beside the first line. `flex-start` fixes it.

## Tests

`FocusOnRouteChangeTests` covers: no focus on initial load, focus after navigating to a different page (now actually varying `RouteData.PageType` between renders, which the original version of this test didn't), no focus when the page re-renders without navigating, and — the case the original test suite didn't cover, which is exactly where the URI-based version of this bug lived — no focus when only the query string changes on the same page.

`dotnet test csharp/PhoneNumbers.Demo.Tests` → 71 passed, 0 failed. The asset move was checked against a real `dotnet publish -c Release -p:PathBase=/libphonenumber-csharp/`: all three SVGs land in the published `wwwroot/`.


twcclegg:
Review notes before handing this to @wilson — one behavioural regression and two places where the description no longer matches the code.

## Focus is lost on the first navigation out of the 404 page

`FocusOnRouteChange` holds `_lastPageType` in component state, but `<NotFound>` doesn't render the component. A 404 therefore disposes it, and the next navigation re-creates it with `_lastPageType == null` — which the wrapper reads as "initial load" and swallows.

Measured on a published build, keyboard navigation, `main` vs this branch:

| step | `main` | this branch |
| --- | --- | --- |
| land on `/does-not-exist`, then → `/parse` | `H1 "Parse & Validate"`, ring | **`BODY`, no focus** |
| `/parse` → `/nope` → `/format` | `H1 "Formatting"`, ring | **`BODY`, no focus** |
| the navigation after that | focused | focused |

It reproduces at unit level too: dispose the component, render a fresh instance, and `Blazor._internal.domWrapper.focusBySelector` is never invoked.

Moving just the initial-load latch into a scoped service (scoped == app lifetime under WASM, so it survives the NotFound round-trip) fixes it while keeping the page-type comparison component-local:

```csharp
if (!Gate.Passed)                        // the app's very first routed render
{
    Gate.Passed = true;
    _lastPageType = RouteData.PageType;
    return;
}

if (_lastPageType is null || _lastPageType != RouteData.PageType)
{
    _lastPageType = RouteData.PageType;  // null => fresh instance after NotFound
    _mount = true;
}
```

That preserves all four cases the current tests cover and restores the 404 case.

## Two stale claims in the description

- **§3** says this adds a `sidebar__link-icon--filled` modifier "since brand marks are filled shapes and the surrounding icons are stroke icons". The shipped code does the opposite — 42a67ef redraws the mark as line art behind `sidebar__link-icon--nuget`. The description argues for a decision that was reversed.
- **§2** and the `index.html` comment say `favicon.svg`'s `prefers-color-scheme` query makes first paint right "before any script runs". On the demo it never runs: the inline `syncFavicon` rewrites `link.href` before the browser fetches an icon, and a network trace of a cold load shows exactly one icon request, `logo.svg`. The claim holds for the docfx site in #453, not for this one.

## Worth considering

- **The NuGet mark renders a visibly lighter stroke than its neighbours.** `viewBox="-1.5 -1.5 27 27"` scales the CSS `stroke-width: 1.75` down to **1.23px** on screen, where every other sidebar icon renders at **1.58px** — 22% thinner, which is why it reads lighter sitting under GitHub. Ink size and centring are fine (measured: dead-centre, and its ink box is actually the largest of the set). `stroke-width: 2.25` would match at 1.58px and still leave the punch-hole open at roughly 1.9px. The CSS comment explains the inset viewBox but not that the inset also shrinks the stroke, which is what makes 1.75 look like a near-match to 2 when it isn't.

- **No CI watches `assets/brand/**`.** `build_and_run_demo_tests.yml` and `deploy-demo.yml` are both path-filtered, and the demo build now consumes those files through the new `Content`/`Link` item. An asset-only change would neither run the demo check nor redeploy the site — the weekly cron would eventually pick the deploy up, but the PR check gap is immediate. One line in each `paths:` list. (#453 will want the same for the docfx workflow.)

- **The tests assert on the marker element, not on focus.** `Assert.Contains("blazor-focus-on-navigate", …)` proves the child component mounted. bUnit can assert the actual behaviour, and the invocation count also covers repeat navigation, which nothing currently does:

  ```csharp
  var handler = JSInterop.SetupVoid("Blazor._internal.domWrapper.focusBySelector", "h1");
  // initial 0 -> first nav 1 -> same-page re-render 1 -> second nav 2
  ```

  Run against this branch that gives exactly those counts. It also silences a latent issue: `JSInterop.Mode` is `Strict` in this project, so the current tests let that invocation go through unplanned.

## Verified working

Repeat navigations are fine — `FocusOnNavigate` re-arms on every `PageType` change, so leaving it permanently mounted after the first navigation does not focus only once (checked against six consecutive navigations). Keyboard nav lands on the new `h1` with the ring, mouse nav focuses without one, initial load leaves focus on `BODY`, and the Home field's query-string sync no longer steals focus. All three SVGs publish into `wwwroot/` under `-p:PathBase=/libphonenumber-csharp/` with the base href rewritten, and the blobs are byte-identical to #453's. The install-block prompt sits on the first line of a wrapped command at 430px. Merges clean with current `main`; demo builds with 0 warnings and 71/71 tests pass on the merged tree.

wmundev (approved):
nice catch on the focus ring, didn't see that before, looks good

twcclegg:
Rebased onto main now that #453 has landed, and folded in the fix for how the two PRs collided. Force-pushed `36d4a2c7` → `f2b19d44`; the nine original commits are unchanged apart from one, described below.

**The brand marks.** #453 moved `assets/brand/` to `docfx/brand/` after review pointed out that nothing outside docfx read it. That was true of main at the time — but this is the branch that makes the demo read it, via the `Content` item in `PhoneNumbers.Demo.csproj` and the `logo.svg`/`logo-dark.svg` swap in `index.html`. Merged as they stood, the two produced **two byte-identical copies** of the three SVGs and no single source of truth, and git merged it with **no conflict**, so nothing would have flagged it.

New commit `f2b19d44` consolidates back onto `assets/brand/` the way both PRs intended: drops `docfx/brand/`, repoints `docfx.json` at `../assets/brand`, restores the `assets/brand/**` path filters to `deploy-demo.yml` and `docs_preview.yml` (without them, editing the marks wouldn't rebuild either site), and corrects the `docfx.json` and `AGENTS.md` comments.

**Generated files.** `9c6e5490` was a one-line `install-block.css` fix that had also swept in a `git add -A` of a local docs build — 309 files, ~62k lines of `docfx/_site/`, `docfx/api/` and `docfx/articles/*.md`. All of it is gitignored on main, but those rules arrived *in* #453, after this branch was cut, so they'd have landed in main permanently (gitignore doesn't untrack). Replayed that commit with just its real change; `docfx/articles/toc.yml` is hand-maintained source and is untouched.

**Verified:** `dotnet test csharp/PhoneNumbers.Demo.Tests` → 98 passed. `./docfx/build.sh` clean at 0 warnings with all three SVGs in `_site/images/`. `dotnet publish -c Release -p:PathBase=/libphonenumber-csharp/` lands all three in the demo's `wwwroot/`. Net diff against main is 19 files, zero generated.

@wmundev — your approval predates the rebase, so a re-look at `f2b19d44` would be worth it if you have a minute.

wmundev (approved):
looks good to me

PR #472, opened 2026-09-15
docs: break down AGENTS.md into various useful reusable skills to follow best practice

wmundev:
Follow best practices from https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices and avoid bloating agents.md / claude.md context


.claude/skills/changing-demo-ui/SKILL.md
```diff
@@ -0,0 +1,105 @@
...
+(singular); overriding the plural property from the command line stops the
+`xunit.runner.visualstudio` build assets from being applied, the xunit discoverer never reaches the
+output directory, and `dotnet test` exits 0 having run nothing. Output ending in
+`No test is available in …` is a failure, not a pass — the same project reports dozens of tests
+without the switch.
+
+## 4. Verify in the browser
+
+Start the preview with the `demo` configuration in `.claude/launch.json` (it runs
+`dotnet run --project csharp/PhoneNumbers.Demo` on <http://localhost:5099>). Then:
```
twcclegg: **Blocking: `.claude/launch.json` and port 5099 do not exist.**

This file has never existed in the repo — `git log --all --diff-filter=A -- '*launch.json'` returns nothing, and the only `.claude/` content anywhere is this PR's `skills/`. There is also no `Properties/launchSettings.json` in the demo project, and the string `5099` appears nowhere outside `resources/` data files.

Since step 4 is a mandatory item in the checklist at the top, an agent following this hits a dead end and then improvises a port number.

Either add the launch config as part of this PR (and pin the port in a demo launch profile so the number is real), or drop to:

> Run `dotnet run --project csharp/PhoneNumbers.Demo` and open the URL the SDK prints.

csharp/PhoneNumbers.Demo/AGENTS.md
```diff
@@ -1,193 +1,36 @@
...
-dotnet run --project csharp/PhoneNumbers.Demo
-```
-
-Publish for GitHub Pages (static WASM output):
+## Build, run, test
 
 ```bash
-dotnet publish csharp/PhoneNumbers.Demo -c Release
+dotnet build csharp/PhoneNumbers.Demo
+dotnet run --project csharp/PhoneNumbers.Demo          # or the "demo" entry in .claude/launch.json
```
twcclegg: Same `.claude/launch.json` reference as in `changing-demo-ui/SKILL.md` — the file does not exist in the repo. Both references need the same fix.

.claude/skills/changing-ci-workflows/SKILL.md
```diff
@@ -0,0 +1,81 @@
...
+  ```
+
+  Never `@v7`, never `@main`. Dependabot bumps these monthly and keeps the comment in sync; let it
+  do the bumping rather than hand-editing SHAs.
+- **Every workflow declares top-level `permissions:`**, normally `contents: read`. Escalate at the
+  *job* level only, only to what that job needs, and say why in a comment. Every workflow here has
+  a top-level block — don't add the first one without.
+- **`persist-credentials: false` on every `actions/checkout`** unless the job genuinely pushes
+  (today only the metadata sync does, with `BOT_ACCESS_TOKEN`).
+- **Default to `runs-on: ubuntu-24.04-arm`.** Four workflows run on x64 `ubuntu-latest` — the
```
twcclegg: It is five, not four — `codeql.yml` is missing from the list:

```yaml
runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
```

That resolves to `ubuntu-latest` for the only language in the matrix. The follow-on "only the last documents why" is then also wrong for CodeQL, which carries no runner-choice comment at all.

.claude/skills/changing-ci-workflows/SKILL.md
```diff
@@ -0,0 +1,81 @@
...
+  metadata sync, `finalize_metadata_release`, `scorecard` and `triage_metadata_issues` — and only
+  the last documents why (the Copilot CLI's npm package lacks reliable arm64 binaries). A new x64
+  job should carry a reason in a comment. **There are no Windows or macOS runners** (bar CodeQL's
+  swift matrix leg), so never write a step that only works on Windows.
```
twcclegg: There is no swift matrix leg. `codeql.yml`'s matrix contains a single entry:

```yaml
matrix:
  include:
    - language: csharp
      build-mode: manual
```

The `macos-latest` branch is dead residue from GitHub's starter template and never runs. So the parenthetical invents an exception to a hard rule — and it contradicts the new `AGENTS.md:85`, which states flatly that there are no Windows or macOS runners. Suggest dropping the parenthetical; the unqualified claim is the true one.

twcclegg (commented):
## Review

I checked essentially every concrete claim in the 924 added lines against the tree at `260f4540`. Structurally this is good: valid frontmatter on all nine skills, sensible task decomposition, and it quietly fixes a stale fact the old `AGENTS.md` carried (the MSBuild targets are `Generate*`, not `Build*`). Most claims hold up. The problems cluster in the material that is **new** rather than moved — which is where fabrications tend to creep into a restructure PR.

One blocking item (inline, on `changing-demo-ui`), four cheap corrections, and one piece of dropped content worth recovering.

### Dropped with no new home: the SDK pin and the determinism guard

The old `AGENTS.md` said "`global.json` pins the SDK to 10.0.100 with `latestFeature` roll-forward, and CI verifies the build is reproducible." That survives nowhere now — not in the new `AGENTS.md`, not in any skill, not in `CONTRIBUTING.md`. But `build_and_run_unit_tests_linux.yml:88-112` rebuilds the solution and requires byte-identical assemblies, so this is a rule an agent can break (any build property that emits a timestamp or absolute path) for an opaque CI failure with nothing to point at. `changing-ci-workflows` is the natural home.

Also dropped without a home, each individually minor but worth being a deliberate call rather than an accident: the `LocaleNames`/`LocaleData` split, `AreaCodeMapStorageStrategy`/`DefaultMapStorage`/`FlyweightMapStorage`, `LangVersion` in `Directory.Build.props`, and the Codecov coverage workflow.

### One structural note

Duplication is now the drift surface: the metadata-regex rule is stated three times (`AGENTS.md`, `tuning-hot-paths`, `porting-upstream-changes`) and the public-member sign-off rule twice, at different lengths each time. That is the right call for skills that have to stand alone — but the sign-off rule is the repo's one break-glass rule, so it is worth keeping its two statements verbatim-identical so they cannot drift apart.

### What checks out

Frontmatter: all nine skills parse as YAML, `name` matches the directory, descriptions run 333–517 chars in the "what + when to use" form, bodies are 75–105 lines, and the one long digression is split into `reference/` and linked rather than inlined. Hygiene is clean — LF, UTF-8, final newlines, no trailing whitespace, nothing outside docs touched.

Verified accurate against the tree: the `Generate*` target names and their sentinel `Outputs` (`PhoneNumberMetadata_US`, `en.1`), all seven MetadataBuilder subcommands, the five `LogicalName` prefixes, all seven benchmark filters and `--retained-memory`, the p<0.001 / 20% floor / 17% spread numbers, exit codes 2/3/4 and `EXPECTED_MAJOR_VERSION` default 9, every flag and workflow-input name, the daily cron, `metadata-update/*`, the `changelog-run` marker format, the `libphonenumber-csharp-bot` login gate, all three path-filter lists, 100% SHA pinning, top-level `permissions` on every workflow, `persist-credentials: false` on every checkout but the sync's, the Dependabot shape, `PackageValidationBaselineVersion` 9.0.37 in both csprojs, `AnalysisMode` only in Extensions, CS1591 in the repo `NoWarn`, the coverlet exclusions, the FsCheck property names and the `using FsCheck;` Scorecard note, the fuzz repro command, `BunitContext` + snake_case naming, the 860px breakpoint, and `UrlState.cs`.

And one genuine improvement beyond the restructure: the divergence pointer moves from `csharp/README.md` "Known Issues" to `docs/api-differences-from-java.md`, which is the fuller and more accurate document.

Happy to merge once the `.claude/launch.json` reference is resolved; the rest are cheap enough to fold into the same push.

AGENTS.md
```diff
@@ -1,110 +1,103 @@
...
+- `PhoneNumberUtil` is a partial class split by TFM: `PhoneNumberUtil.net.cs` (modern BCL) and
+  `PhoneNumberUtil.netstandard.cs` (fallbacks). Every public signature must exist on all three TFMs.
+- Subsystems mirror Java types of the same name: `AsYouTypeFormatter`, `PhoneNumberMatcher`,
+  `ShortNumberInfo`, `PhoneNumberOfflineGeocoder` / `PhoneNumberToCarrierMapper` /
+  `PhoneNumberToTimeZonesMapper` (backed by `AreaCodeMap` prefix maps).
+- Parsing and formatting are allocation-light on purpose: spans and slices over substrings and
+  `Match` objects; lookup tables built once into frozen collections. Measure hot-path changes with
+  the benchmark harness rather than reasoning about them.
+- Nullable reference types are on everywhere except `netstandard2.0`; annotate new code regardless.
+- CI is GitHub Actions on `ubuntu-24.04-arm` only — no Windows or macOS runners. Releases are
```
twcclegg: "on `ubuntu-24.04-arm` only" is contradicted by `changing-ci-workflows`, which correctly notes that several workflows run on `ubuntu-latest` (five of them, counting CodeQL).

"The default runner is `ubuntu-24.04-arm`; a handful of jobs run x64 on `ubuntu-latest`" would be accurate and keeps the useful half of the rule — that there are no Windows or macOS runners.

.claude/skills/changing-ci-workflows/SKILL.md
```diff
@@ -0,0 +1,81 @@
...
+## CI tooling: bash or C#, never JavaScript
+
+Scripts under `lib/` are bash (`set -euo pipefail`, `jq` for JSON). Anything needing real data
+structures, statistics or a library is a small C# console project like
+`csharp/PhoneNumbers.BenchmarkTools/`, run with `dotnet run --project` and kept out of the
+solution. A few `lib/*.js` helpers once crept in as an implementation detail and were ported away;
+JavaScript belongs only in the Blazor demo's own web assets.
+
+Shared shell functions live in `lib/github-release-helpers.sh`, which is sourced rather than run.
+Keep scripts bash-3.2-compatible (no `${var,,}`), since maintainers run them on macOS.
```
twcclegg: This is scoped repo-wide but the repo does not hold to it. Only `lib/github-release-helpers.sh:9` carries the bash-3.2 note, and `lib/update-changelog.sh:59` already uses `mapfile`, which is bash 4+ and unavailable on macOS's stock bash.

As written an agent will either "fix" that `mapfile` line or trust a portability guarantee that isn't there. Worth either scoping the rule to the sourced helper, or stating the existing exception explicitly.

.claude/skills/changing-ci-workflows/SKILL.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Dependabot
+
+`.github/dependabot.yml` covers two ecosystems: `github-actions` at `/`, and `nuget` at `/csharp` —
+one entry, because every version lives in `csharp/Directory.Packages.props`. Major updates are
```
twcclegg: Small precision point: the `ignore` block for `version-update:semver-major` is on the **nuget** entry only. GitHub Actions majors are not ignored — Dependabot does bump those.

.claude/skills/building-embedded-metadata/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Building the embedded metadata
+
+Nothing under `resources/` ships as-is and nothing is read from disk at run time.
+`PhoneNumbers.MetadataBuilder` converts the XML and text files into per-region binaries during
+`dotnet build`, and `PhoneNumbers.csproj` embeds them gzip-compressed into the assembly. At run time
+`MetadataSource` + `EmbeddedResourceMetadataLoader` pull a region's binary out of the assembly's
+resources on first use.
+
+The whole pipeline lives in **`csharp/PhoneNumbers/PhoneNumbers.csproj`, roughly lines 50–220**.
```
twcclegg: The targets actually span lines 109–233 (`GenerateBinaryMetadata` at 109 through `CleanBinaryMetadata` ending at 233). Line ranges in docs rot on the first edit to the csproj, and the table right below already names every target — probably better to drop the range entirely.

.claude/skills/building-embedded-metadata/SKILL.md
```diff
@@ -0,0 +1,83 @@
+---
+name: building-embedded-metadata
+description: Understand or change how resources/ becomes the binary metadata embedded in the assembly — the PhoneNumbers.MetadataBuilder tool, the MSBuild Generate*/Embed targets in PhoneNumbers.csproj, and the loaders that read the result. Use when metadata seems stale or missing after a local edit, when a build fails or races in those targets (CS2012, parallel writes to obj, MissingMetadataException), or when adding a new kind of embedded data.
+---
+
+# Building the embedded metadata
+
+Nothing under `resources/` ships as-is and nothing is read from disk at run time.
+`PhoneNumbers.MetadataBuilder` converts the XML and text files into per-region binaries during
+`dotnet build`, and `PhoneNumbers.csproj` embeds them gzip-compressed into the assembly. At run time
```
twcclegg: Minor attribution slip: MetadataBuilder does the gzip when it writes each bin (`new GZipStream(File.Create(outPath), CompressionLevel.SmallestSize)` in `Program.cs`). `PhoneNumbers.csproj` embeds files that are already compressed. Worth getting right, since someone debugging a `MissingMetadataException` will go looking for the compression step.

.claude/skills/changing-demo-ui/SKILL.md
```diff
@@ -0,0 +1,105 @@
...
+- Layout is sidebar + main; at `max-width: 860px` the sidebar collapses behind a toggle and the
+  grids (`input-row`, `result-grid`, `feature-grid`) go single-column.
+- Light theme only today. If adding dark mode: light defaults in custom properties, overrides under
+  `[data-theme="dark"]`, contrast verified in both.
+
+## 2. Tests: what to assert
+
+bUnit, xUnit, classes derive from `BunitContext`, one file per page in
+`csharp/PhoneNumbers.Demo.Tests/Pages/`, test names in snake_case describing behaviour
+(`shows_valid_badge_for_default_uk_number`). Arrange-Act-Assert; one concept per test.
```
twcclegg: The prepopulated number on the home page is US, not UK — the real test is `shows_valid_badge_for_prepopulated_us_number`. Worth matching, since the example is otherwise close enough to a real name to be read as one.

.claude/skills/changing-demo-ui/SKILL.md
wmundev: Removed. It now says to run `dotnet run --project csharp/PhoneNumbers.Demo` and open the URL the SDK prints.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/.claude/skills/changing-demo-ui/SKILL.md#L73-L78

csharp/PhoneNumbers.Demo/AGENTS.md
wmundev: Removed here too.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/csharp/PhoneNumbers.Demo/AGENTS.md#L15

.claude/skills/changing-ci-workflows/SKILL.md
wmundev: Fixed to five, with codeql added.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/.claude/skills/changing-ci-workflows/SKILL.md#L28-L33

.claude/skills/changing-ci-workflows/SKILL.md
wmundev: Dropped the parenthetical. It now just says there are no Windows or macOS runners.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/.claude/skills/changing-ci-workflows/SKILL.md#L32-L33

AGENTS.md
wmundev: Used your wording.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/AGENTS.md#L91-L93

.claude/skills/changing-ci-workflows/SKILL.md
wmundev: Rule now applies only to the helper script, and it calls out the `mapfile` line so nobody "fixes" it.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/.claude/skills/changing-ci-workflows/SKILL.md#L46-L49

.claude/skills/changing-ci-workflows/SKILL.md
wmundev: Fixed.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/.claude/skills/changing-ci-workflows/SKILL.md#L53-L57

.claude/skills/building-embedded-metadata/SKILL.md
wmundev: Dropped the line range.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/.claude/skills/building-embedded-metadata/SKILL.md#L15-L17

.claude/skills/building-embedded-metadata/SKILL.md
wmundev: Fixed: MetadataBuilder does the gzip, the csproj embeds the result.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/.claude/skills/building-embedded-metadata/SKILL.md#L8-L13

.claude/skills/changing-demo-ui/SKILL.md
wmundev: Fixed to the real name.
https://github.com/twcclegg/libphonenumber-csharp/blob/bbcb00e8/.claude/skills/changing-demo-ui/SKILL.md#L44

.gitignore
```diff
@@ -199,8 +199,11 @@ FakesAssemblies/
 .DS_Store
 .idea
 
-# Claude Code local worktrees (background agent scratch state)
-.claude/
+# Claude Code: worktrees are background-agent scratch state; settings.local.json and
+# launch.json are per-developer. Everything else under .claude/ (skills) is shared repo content.
+.claude/worktrees/
+.claude/settings.local.json
+.claude/launch.json
```
twcclegg: The three paths are right — `ff00f5c9`'s own commit message confirms `.claude/worktrees/` is where the background-agent worktrees land, so this narrowing is well-founded.

The concern is the shape rather than the contents. The broad ignore came from `7aa205f5`, whose rationale was "Claude Code's own local state, **including** per-agent git worktrees" — deliberately broad, because an enumeration can't anticipate what the tool writes there next. An allow-list keeps the intent while still tracking the skills:

```gitignore
.claude/*
!.claude/skills/
```

Separately, `.claude/launch.json` on this line is now vestigial — nothing references it since the fix, so it's ignoring a file that no longer has a role.

.claude/skills/changing-public-api/SKILL.md
```diff
@@ -0,0 +1,77 @@
...
+# Changing the public surface
+
+Both packable projects have `EnablePackageValidation` on, so the build itself is the gate: an API
+that is inconsistent across target frameworks, or a breaking change against the published baseline,
+fails the build rather than shipping.
+
+## First: is a new member authorised?
+
+**Adding a public member to `csharp/PhoneNumbers/` needs explicit sign-off from the user, as its
+own decision.** Package validation only catches removals, so nothing automated will object. "It
```
twcclegg: The two copies of this rule are now byte-identical with `AGENTS.md:57`, which is what I asked for — but they got there by shortening the more accurate one, and it now contradicts line 38 of this same file:

> **TFM parity is mandatory.** A member that exists on `net10.0` but not `netstandard2.0` fails package validation.

So package validation catches more than removals. The point being made here — that *additions* get no automated pushback — is the part that matters and it survives either phrasing. Suggest making both copies:

> Package validation only catches breaks against the published baseline — never additions, so nothing automated will object.

.claude/skills/changing-ci-workflows/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+## Dependabot
+
+`.github/dependabot.yml` covers two ecosystems: `github-actions` at `/`, and `nuget` at `/csharp` —
+one entry, because every version lives in `csharp/Directory.Packages.props`. NuGet major updates
+are ignored deliberately and minor/patch NuGet updates are grouped; GitHub Actions majors are not
+ignored, and `github/codeql-action*` is grouped because the CodeQL Action refuses to run with
+mismatched `init`/`analyze` versions. Adding a package
+manifest outside `/csharp` means adding an entry; adding another project inside it does not.
+
+## The build must be reproducible
```
twcclegg, line 60: This section is accurate and well written — my only worry is that the people who need it won't find it.

It lives in a skill whose `description:` is entirely about `.github/`, actions, runners, permissions, Scorecard and `lib/` tooling. But the rule actually fires when someone edits `csharp/Directory.Build.props` — adding a build property that embeds a timestamp or an absolute path — which reads as a build-settings change, not a CI change. An agent reaching for a skill on that task would not open `changing-ci-workflows`.

Either add a clause to this skill's `description:` (something like "or when adding an MSBuild property that could make the build non-deterministic"), or promote a one-liner into `AGENTS.md`'s **Hard rules**. Right now the only mention there is buried mid-sentence in an Architecture bullet.

AGENTS.md
```diff
@@ -1,112 +1,112 @@
...
+  `ShortNumberInfo`, `PhoneNumberOfflineGeocoder` / `PhoneNumberToCarrierMapper` /
+  `PhoneNumberToTimeZonesMapper` (backed by `AreaCodeMap` prefix maps, stored via
+  `AreaCodeMapStorageStrategy` — `DefaultMapStorage` or `FlyweightMapStorage`).
+- Parsing and formatting are allocation-light on purpose: spans and slices over substrings and
+  `Match` objects; lookup tables built once into frozen collections. Measure hot-path changes with
+  the benchmark harness rather than reasoning about them.
+- Nullable reference types are on everywhere except `netstandard2.0`; annotate new code regardless.
+- CI is GitHub Actions. The default runner is `ubuntu-24.04-arm`; a handful of jobs run x64 on
+  `ubuntu-latest`; there are no Windows or macOS runners. PRs run `build_and_run_unit_tests_linux.yml`
+  (net10.0), `run_all_tests_and_upload_code_coverage.yml` (every TFM, Codecov) and `codeql.yml`, and the unit-test job rebuilds the packages and
```
twcclegg: Cosmetic: this line is 144 characters where the rest of the file wraps at ~100. Worth a reflow since the wrapping is otherwise consistent throughout.

twcclegg (approved):
## Re-review — all ten findings fixed, approving

Re-checked at `bbcb00e8` against the tree rather than taking the diff at face value. Every one of the ten points is addressed, and addressed correctly.

A few are better than what I asked for. The bash-3.2 rewrite not only scopes the rule to `github-release-helpers.sh` and names the `mapfile` exception, it adds a "don't add bash-4 constructs to the helper either" guard that the original didn't have. The reproducibility section was the one I most expected to drift, and it doesn't: "packs, cleans, packs again, fails if any `PhoneNumbers.dll` differs by hash — for every TFM, and after a clean that also regenerates the metadata bins" matches `build_and_run_unit_tests_linux.yml:88-112` exactly, down to the netstandard2.0/net8.0/net10.0 loop and the `sha256sum` comparison, and `ContinuousIntegrationBuild`-under-`CI` plus `EmbedUntrackedSources` really are at `Directory.Build.props:38-39`.

### The rebase is the part worth calling out

`main` changed the sync semantics underneath this PR while it was open (#464 moved auto-merge from "on at open" to "off at open, armed as a backstop by a later run"). Stale skill text describing the old flow would have been worse than no skill text at all. That got caught: `syncing-upstream-metadata` is rewritten for the two-phase flow, and the reference doc grew two new sections for it.

I traced every non-obvious claim in that new material to a source rather than spot-checking. "GitHub only accepts `enablePullRequestAutoMerge` on a PR that is *blocked*… 'Pull request is in clean status'" is verbatim from `lib/github-release-helpers.sh:72-73`. "A failure to arm auto-merge on the backstop run is fatal, not a warning" matches the `fail 1` at `github-actions-metadata-update.sh:524` and its comment. "`metadata-update/*` sits outside every ruleset", "upstream releases are at least five days apart", and "guards… were written and then removed" all come from #464's commit message and PR body. Nothing invented — including the two new "Don't push fixes onto the branch" / "Closing the PR does not decline the release" warnings, which match the generated PR body at line 498.

The `lib/github-actions-metadata-update.sh` change is comment-only — two doc pointers, no executable line touched — so the scope growth to 15 files is fine.

### Structure still clean

All nine skills: valid YAML frontmatter, `name` matches directory, exactly `name` + `description`, descriptions 333–517 chars, bodies 75–111 lines. The skills table in `AGENTS.md` matches the directory listing. LF, UTF-8, final newlines, no trailing whitespace anywhere in the diff.

### Four minor things, inline

None worth blocking on — the two one-line edits are worth folding in if you're pushing again anyway, the other two are fine as follow-ups. Approving on the understanding that they're your call.

Nice work on this one. The restructure holds up under a second pass, which is not the usual outcome for a docs PR this size.

.gitignore
wmundev: Switched to the allow-list and dropped `launch.json`.
https://github.com/twcclegg/libphonenumber-csharp/blob/d12d4480/.gitignore#L202-L206

.claude/skills/changing-public-api/SKILL.md
wmundev: Used your wording in both copies.
https://github.com/twcclegg/libphonenumber-csharp/blob/d12d4480/AGENTS.md#L56-L61
https://github.com/twcclegg/libphonenumber-csharp/blob/d12d4480/.claude/skills/changing-public-api/SKILL.md#L14-L19

.claude/skills/changing-ci-workflows/SKILL.md
wmundev: Did both: added the MSBuild-property clause to the description, and promoted a one-liner into Hard rules.
https://github.com/twcclegg/libphonenumber-csharp/blob/d12d4480/.claude/skills/changing-ci-workflows/SKILL.md#L3
https://github.com/twcclegg/libphonenumber-csharp/blob/d12d4480/AGENTS.md#L71-L73

AGENTS.md
wmundev: Reflowed. The reproducibility clause moved to Hard rules, so the bullet just points at it.
https://github.com/twcclegg/libphonenumber-csharp/blob/d12d4480/AGENTS.md#L95-L99

PR #473, opened 2026-09-17
feat: extract svgs in the demo website to reusable components instead of inlining them everywhere

wmundev:
## Extract demo SVG icons into reusable components

The demo had 34 inline `<svg>` blocks pasted across `MainLayout.razor`, `Home.razor` and the four empty-state pages — the same glyph copied up to five times (globe, arrow-right), and the theme toggle built its icon from an SVG string via `AddMarkupContent`. This PR replaces them with 18 Razor icon components and makes "no inline SVG" a documented rule.

### Changes

**`Components/Icons/` (new)**
- `IconBase.cs` — abstract base with `CaptureUnmatchedValues`, so callers can pass `class`, `data-*`, etc. through to the `<svg>`.
- `Icon.razor` — the single shared `<svg viewBox="0 0 24 24" aria-hidden="true" focusable="false">` wrapper.
- 18 two-line glyph components (`SearchIcon`, `GlobeIcon`, `ArrowRightIcon`, `SunIcon`, `MoonIcon`, …), used as `<SearchIcon />`. Namespace added to both `_Imports.razor` files.

**Call sites**
- Layout and pages now use the components. `ThemeIcon()` is `_theme == "dark" ? @<SunIcon /> : @<MoonIcon />` instead of a markup string.
- Empty-state SVGs carried `width="48" fill="none" stroke="…"` attributes that the block CSS was already overriding; dropping them is a no-op (verified: still 26px).
- `BoxIcon` (hero badge outline) and `PackageIcon` (sidebar NuGet, with fold lines) stay separate — they are different glyphs and merging them would be a visible change at 14px.

**Tests**
- `PhoneNumbers.Demo.Tests/Components/IconTests.cs` reflects over every `IconBase` subclass and asserts a single `aria-hidden` 24×24 `<svg>` with content, plus one attribute-forwarding test. New icons are covered automatically.
- `dotnet test csharp/PhoneNumbers.Demo.Tests`: 86 passed (67 existing + 19 new).

**Docs / skill**
- `.claude/skills/changing-demo-ui/SKILL.md`: new *"SVG icons are components, never inline markup"* section (naming, reuse-before-add, CSS owns size/stroke, decorative-only a11y contract), a Don'ts entry, and the description mentions the rule so the skill triggers on icon work.
- `PhoneNumbers.Demo/AGENTS.md` rule list and `README.md` project tree updated.

### Verification
Ran the demo preview: all 27 rendered SVGs are in the SVG namespace and `aria-hidden`, no console errors. Theme toggle swaps sun↔moon with the matching `aria-label`; at mobile width the topbar menu/share/theme icons render at 20px and empty-state icons at 26px, identical to before.

No visual or behavioural change intended.


csharp/PhoneNumbers.Demo/Layout/MainLayout.razor
```diff
@@ -220,13 +218,8 @@
...
 
-    private RenderFragment ThemeIcon() => builder =>
-    {
-        // sun when in dark mode (to switch to light), moon otherwise
-        builder.AddMarkupContent(0, _theme == "dark"
-            ? "<svg viewBox=\"0 0 24 24\" aria-hidden=\"true\"><circle cx=\"12\" cy=\"12\" r=\"4\"/><path d=\"M12 2v2M12 20v2M4.93 4.93l1.41 1.41M17.66 17.66l1.41 1.41M2 12h2M20 12h2M4.93 19.07l1.41-1.41M17.66 6.34l1.41-1.41\"/></svg>"
-            : "<svg viewBox=\"0 0 24 24\" aria-hidden=\"true\"><path d=\"M21 12.79A9 9 0 1 1 11.21 3 7 7 0 0 0 21 12.79z\"/></svg>");
-    };
+    // sun when in dark mode (to switch to light), moon otherwise
+    private RenderFragment ThemeIcon() => _theme == "dark" ? @<SunIcon /> : @<MoonIcon />;
```
twcclegg, line 222: This is the only behavioural rewrite in the PR, and it's the only thing here with no test behind it.

`MainLayout` has zero bUnit coverage — there's no `MainLayoutTests`, and nothing in the suite touches theme, the share button or the sidebar. So the 19 new tests cover two-line components that essentially cannot break, while the one non-trivial change (markup string → component ternary) rests entirely on manual browser checking.

A small `MainLayoutTests` asserting the toggle swaps sun↔moon and that `aria-label` follows the state would guard the thing that can actually regress. The gap predates this PR — this is just the commit that makes it matter.

csharp/PhoneNumbers.Demo.Tests/Components/IconTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        var cut = Render(builder =>
+        {
+            builder.OpenComponent(0, icon);
+            builder.CloseComponent();
+        });
+
+        var svg = Assert.Single(cut.FindAll("svg"));
+        Assert.Equal("true", svg.GetAttribute("aria-hidden"));
+        Assert.Equal("0 0 24 24", svg.GetAttribute("viewBox"));
```
twcclegg, line 30: Worth asserting `focusable="false"` here too. The new skill section documents it as part of the wrapper contract —

> inherits the `aria-hidden`/`focusable="false"`/`viewBox` wrapper from `Icon.razor`

— but nothing enforces it, so it could be dropped from `Icon.razor` without a single test failing. One extra line closes the gap between what's documented and what's checked.

csharp/PhoneNumbers.Demo.Tests/Components/IconTests.cs
```diff
@@ -0,0 +1,40 @@
+using Bunit;
+using PhoneNumbers.Demo.Components.Icons;
+using Xunit;
+
+namespace PhoneNumbers.Demo.Tests.Components;
```
twcclegg, line 5: This new `Components/` test folder is the right home for these, but three places still describe the old layout as the only one:

- `csharp/PhoneNumbers.Demo/AGENTS.md:35` — "Every logic change gets a bUnit test in `csharp/PhoneNumbers.Demo.Tests/Pages/`"
- `.claude/skills/changing-demo-ui/SKILL.md:16` — checklist step 2
- `.claude/skills/changing-demo-ui/SKILL.md:69` — "one file per page in `csharp/PhoneNumbers.Demo.Tests/Pages/`"

The PR updates the icon rules carefully and then leaves the test-location sentences pointing at a layout it just changed. Something like "one file per page in `Pages/`, component tests in `Components/`" covers it.

csharp/PhoneNumbers.Demo/Components/Icons/Icon.razor
```diff
@@ -0,0 +1,9 @@
+@* Shared <svg> wrapper used by every icon component in this folder. Not used directly by pages. *@
+<svg viewBox="0 0 24 24" aria-hidden="true" focusable="false" @attributes="AdditionalAttributes">@ChildContent</svg>
+
+@code {
+    [Parameter] public RenderFragment? ChildContent { get; set; }
+
+    [Parameter(CaptureUnmatchedValues = true)]
+    public Dictionary<string, object>? AdditionalAttributes { get; set; }
```
twcclegg, line 10: I think this deliberately re-declares `AdditionalAttributes` rather than `@inherits IconBase`, because inheriting would make `Icon` a non-abstract `IconBase` subclass — and `IconTests.AllIcons()` would then pick it up, render it with no children, and fail `Assert.NotEmpty(svg.Children)`.

That's a reasonable call, but the reasoning is invisible to the next reader, who may "tidy" it into inheriting and break the suite in a way that looks unrelated. Worth one comment line saying why it stands apart — the comment on line 1 explains what `Icon` is for, but not why it doesn't share the base.

twcclegg (approved):
## Review — approving

I went looking specifically for the three ways this refactor usually goes wrong: a glyph quietly substituted for a similar one, a dropped presentation attribute the CSS doesn't actually replace, and the SVG-namespace trap. None of them are here, and the numbers in the description all reconcile.

### What I checked rather than took on trust

**Glyph-for-glyph equivalence across all 34 call sites.** I reconstructed each changed `.razor` from `main`, pulled every inline `<svg>`'s inner markup in document order, and compared it against the component that replaced it at that position:

```
MainLayout.razor: 16 icons, all match     GeoTimezone.razor:  2 icons, all match
Home.razor:       13 icons, all match     FindNumbers / Formatting / ParseValidate: 1 each, all match
                                           checked 34 call sites, 0 mismatches
```

Whitespace-normalised byte equality, including the two SVGs that were previously escaped inside the `AddMarkupContent` string.

**The dropped presentation attributes really are no-ops.** Two groups were removed, and the description only justifies one:

| Removed | Covered by |
| --- | --- |
| Empty states x5: `width`/`height="48" fill="none" stroke="currentColor" stroke-width="1.5"` | `.empty-state__icon svg` sets all five, `stroke-width: 1.5` included |
| Sidebar logo x1: `fill="white" width="20" height="20"` | `.sidebar__logo-icon svg { width:20px; height:20px; fill:#fff }` |

The logo one isn't mentioned anywhere in the PR body, and `fill="white"` disappearing is exactly what a reviewer would stop on. It is genuinely a no-op — worth a line in the description so the next person doesn't have to re-derive it. The other 28 sites had no fill/stroke attributes to begin with, so they were already CSS-driven and can't have regressed.

**The SVG namespace is safe by construction, not by luck.** This is the real trap in this refactor: the `<svg>` is now created by `Icon.razor` while `<circle>`/`<line>` arrive as a *different* component's `RenderFragment`, and bUnit renders through AngleSharp so it cannot catch a namespace bug. From the shipped `blazor.webassembly.js`:

```js
insertElement(...) { const l = c.elementName(s),
  u = "svg" === l || oe(n) ? document.createElementNS("http://www.w3.org/2000/svg", l) : ...

function oe(e){ const t=le(e); return "http://www.w3.org/2000/svg"===t.namespaceURI && "foreignObject"!==t.tagName }
```

The decision is made from the **DOM parent**, not the component boundary, so children inherit the namespace no matter which component emitted them. Your browser check confirmed it empirically; this is why it can't drift later.

**Build, tests, trimmer.** `dotnet test csharp/PhoneNumbers.Demo.Tests` → 86 passed, matching the claimed 67 + 19. `dotnet publish -c Release` clean with zero trim/AOT warnings — the reflection lives only in the test project, so nothing reaches the WASM output. Hygiene clean throughout: LF, final newlines, no trailing whitespace.

**Counts reconcile.** 18 glyphs + `Icon.razor` + `IconBase.cs`; every component used at least once, no dead ones; 34 call sites = 34 replaced SVGs; no inline `<svg>` or `AddMarkupContent` left in `Pages/` or `Layout/`, so the PR complies with the rule it introduces. The "27 rendered" figure checks out too (11 sidebar + 3 topbar + 3 hero badges + 10 feature grid, conditional pairs counted once).

**`FileIcon`/`FileTextIcon` and `BoxIcon`/`PackageIcon` are not duplicates.** I pairwise-compared all 18 for identical glyph markup — zero duplicates. `FileText` adds two content lines to `File`, `Package` adds fold lines to `Box`, matching feather's own naming. The description defends the Box/Package split but not File/FileText; both are correct.

### Notes, not findings

`aria-hidden="true"` is newly added to ~18 icons that lacked it. I checked each one: every one sits inside a wrapper that already carries `aria-hidden="true"`, or is a badge followed by visible text. Redundant, never a regression. `focusable="false"` is likewise new on all 34 — a small improvement, though it does make "no behavioural change intended" very slightly untrue.

Four minor things inline. Only the first is one I'd actually like to see, and it works fine as a follow-up.

csharp/PhoneNumbers.Demo/Layout/MainLayout.razor
wmundev: Fixed: https://github.com/twcclegg/libphonenumber-csharp/blob/58a5fdb3/csharp/PhoneNumbers.Demo.Tests/Layout/MainLayoutTests.cs

csharp/PhoneNumbers.Demo.Tests/Components/IconTests.cs
wmundev: Fixed: https://github.com/twcclegg/libphonenumber-csharp/blob/58a5fdb3/csharp/PhoneNumbers.Demo.Tests/Components/IconTests.cs#L30

csharp/PhoneNumbers.Demo.Tests/Components/IconTests.cs
wmundev: Fixed: https://github.com/twcclegg/libphonenumber-csharp/commit/58a5fdb3

csharp/PhoneNumbers.Demo/Components/Icons/Icon.razor
wmundev: Fixed: https://github.com/twcclegg/libphonenumber-csharp/blob/58a5fdb3/csharp/PhoneNumbers.Demo/Components/Icons/Icon.razor#L1-L3

PR #474, opened 2026-09-19
feat: let a trimmed build drop the geocoding and country-name data

twcclegg:
## What

Two MSBuild properties let a **trimmed** build drop bundled data sets it does not use:

```xml
<PublishTrimmed>true</PublishTrimmed>
<PhoneNumbersIncludeGeocodingData>false</PhoneNumbersIncludeGeocodingData>
<PhoneNumbersIncludeLocaleNameData>false</PhoneNumbersIncludeLocaleNameData>
```

| trimmed `PhoneNumbers.dll` | bytes |
| --- | --- |
| both properties unset (today's behaviour) | 2,724,864 |
| geocoding dropped | 656,384 |
| both dropped | **259,584** (−90.5%) |

NativeAOT sees the same effect end to end: 5,803,976 → 3,338,216 bytes.

Both default to included, so a project that sets nothing is unchanged, and neither does anything in an untrimmed build.

## Why

The assembly carries ~2.6 MB of embedded data and only 136 KB of it — the phone metadata — is needed to parse, validate and format. The rest backs the offline geocoder, the carrier and time zone mappers, and the country-name table.

Embedded resources survive trimming: ILLink removes code, not resources. Two otherwise identical NativeAOT apps, one calling the geocoder and one not, differ by only 184 KB — both ship the full 1.9 MB of area descriptions.

## No public API change

The switches are internal; the opt-in is an MSBuild property shipped in `buildTransitive/`, which is packaging rather than API. `dotnet pack` reports no API differences against the 9.0.37 baseline.

## Why the data is packed

ILLink matches `<resource>` by exact name and supports no wildcard — `PhoneNumbers.geocoding.*` is reported as not found (IL2040), not expanded. One resource per file would mean a ~690-entry substitutions file regenerated on every fortnightly metadata sync. Each multi-file data set is packed into a single embedded resource (`ResourcePack`), so `ILLink.Substitutions.xml` is **four fixed lines that never need regenerating**.

A pack holds only its directory and reads entries on demand. Materialising them at load is the obvious alternative and costs **1.9 MB of permanently retained memory** for any process that geocodes (1511 KB → 3430 KB; 1529 KB as implemented), because an embedded resource stream is a view over the already-mapped assembly image. The repo's `--retained-memory` audit does not exercise the geocoder, so it does not cover this.

## Build pipeline

- five `Generate*` targets driving **seven** `MetadataBuilder` processes → **one** target, one `all` invocation
- `Inputs`/`Outputs` names its real outputs instead of sentinel files standing in for directories of ~220
- the tool creates its own output directories, so nothing depends on a caller's `<MakeDir>`
- embedded resource count drops from ~690 to ~545

The cross-process mutex in `MetadataBuilder.Main` already handled write races and is untouched.

## Failure behaviour

With a data set removed the geocoder does not fail on its own — it returns `""` and exits 0. Each affected type therefore throws `MissingMetadataException` (an `InvalidOperationException`, and the type the time zone mapper already threw for an absent resource) naming the property to remove.

`PhoneNumbersIncludeLocaleNameData=false` requires `PhoneNumbersIncludeGeocodingData=false`, enforced at build time as **PN1001**: the geocoder falls back to the country name for numbers with no area-level description, so keeping it without the country names geocodes US, DE and GB correctly and throws for Andorra and San Marino.

Property values are normalised before reaching ILLink and a non-boolean is rejected as **PN1002**. MSBuild condition operators coerce booleans and treat `yes/no/on/off` as such, while ILLink accepts only the literal strings — so `no` would otherwise reach it as a malformed feature argument and fail with `IL1030`, naming neither the property nor this package.

## Measured cost

Benchmarked against `origin/main` with the repo's own BenchmarkDotNet harness, both assemblies loaded by the same harness binary.

| | base | this PR |
| --- | --- | --- |
| steady-state geocode (200k warm calls) | 106.31 ms / 19,200,000 B | 106.34 ms / 19,200,000 B |
| `PrefixFileReader` construction | 257 µs / 225 KB | **199 µs / 107 KB** |
| `FirstUseGeocode` allocated | 812.67 KB | **697.71 KB** |
| clean build (3 TFMs) | 3.51 s | **2.78 s** |
| shipped `PhoneNumbers.dll` | 2,864,640 B | **2,837,504 B** |

Warm paths are unchanged or better. `Assembly.GetManifestResourceStream` got *cheaper* because the manifest shrank from 1234 entries to 546 (~3.5 µs → ~1.8 µs per call), and two manifest scans became dictionary lookups.

Three costs, all small and all real:

- **+0.8 ms once per process**, JIT on the first pack open, and only when nothing has already touched `BinaryReader`. `ProcessStartBenchmark` is flat (11.50 → 11.60 ms).
- **+16–30 KB permanently retained** — the pack directories. The `--retained-memory` audit is unchanged at 1219 KB (budget 1600 KB); the delta only appears on paths that touch a pack. Against the 1.9 MB that materialising entries would have cost, this is the intended trade, but it is not literally zero.
- **A new CI step**, two trimmed publishes with a clean restore each. 11.7 s on 32 cores; slower on the 4-core runner, still well inside the 30-minute budget.

One thing I could not explain: `PhoneNumberWorkflowBenchmark.ParseWithExtension` is **+4.4%** with the same sign across 5 paired alternating runs (720.6 → 752.4 µs median), while its allocations are byte-identical and nothing in the diff touches the parse or extension path. Most likely assembly layout shifting with two new source files. It sits below CI's 20% flag floor, so CI will not report it — flagging it here rather than leaving it silent.

## Verification

- clean build of all three TFMs, no warnings; **1018 tests** pass on net8.0 and net10.0
- `dotnet pack`: no API-compat errors; packing twice gives byte-identical assemblies on all three TFMs
- consumed as a real NuGet package (not a ProjectReference) with both properties set
- trimmed and NativeAOT publishes both drop the data with zero IL warnings
- an untrimmed build with both properties set keeps every data set and works normally
- trimmed Blazor demo publishes clean
- from-scratch build reproduces the same pack bytes as an incremental one; `dotnet clean` still removes the generated dirs
- a simulated metadata sync regenerates the pack and reverts to identical bytes
- property values `false` / `False` / `no` / `off` all drop the data; `banana` is PN1002; `true`+`false` is PN1001

`TestResourcePack.cs` covers the container (round-trip, determinism regardless of input order, repeat reads, truncation, implausible entry count, duplicate names) plus two drift guards: every name in `ILLink.Substitutions.xml` must be a real embedded resource, and each removable data set must be exactly one resource. Both were checked to fail when a resource is renamed or the substitutions file is unembedded.

## Alternatives measured and rejected

- **`ZipArchive`** would have replaced `ResourcePack` entirely — named entries, per-entry deflate, `GetEntry(name).Open()` as a bounded sub-stream. Measured: the data is 31 KB larger as zip entries, and referencing it adds ~66 KB of trimmed IL (`System.IO.Compression` 34 KB → 100 KB). That is ~97 KB on an opted-out `PhoneNumbers.dll` of ~259 KB, charged to consumers who never opt out. Recorded in the `ResourcePack` class doc.
- **`System.IO.Packaging`** — an extra NuGet dependency on a library whose only references are netstandard polyfills. **`ResourceManager`/`.resources`** — no compression, and a known trim/AOT liability under `IsAotCompatible`.
- **Three or four properties** instead of two: carrier (108,713 B) and timezones (12,312 B) together are under 6% of what `PhoneNumbersIncludeGeocodingData` covers (2.07 MB), for two more knobs.
- **Separate NuGet packages**: `PhoneNumberOfflineGeocoder` is existing public API of the main package, so a split means relocating types and type-forwarding, against an additive and reversible MSBuild property.

## Follow-ups worth recording, not done here

- **One gzip stream per data set instead of per entry.** Measured: locale would be 241,960 B rather than 397,127 B — **−155 KB, −39%** — because its entries are small and highly similar. Per-entry gzip is clearly right for geocoding (whole-set compression would mean decompressing 8.8 MB to read one map), but locale is worth revisiting.
- **Pack the 541 per-region metadata resources too.** Only 135,850 B of payload but ~541 manifest entries of overhead, and it would retire the last sentinel `Outputs`. Left alone because it touches the cold-start metadata loader, which this change otherwise does not.
- **A shared magic/version header helper.** `ResourcePack` is the fifth copy of the read-magic/check-version shape; the other four pre-date this PR. `BuildPrefixMapFromBin` deliberately collects its magics in one place, and this breaks that.

## Open questions

1. **Naming.** `PhoneNumbersIncludeGeocodingData` also drops carrier and time zone data. `IncludePrefixMapData` would be more accurate and matches the internal vocabulary (`PrefixFileReader`, `BuildPrefixMapFromBin`) but is less discoverable for someone searching "geocoding". Current choice favours discoverability, with diagnostics that enumerate the full set.
2. **Splitting.** Three flags (geocoding / carrier / time zones) were considered and rejected: carrier is 106 KB and time zones 12 KB against geocoding's 1.9 MB, so it triples the combinations for a rounding error.
3. **Packing the metadata resources too.** They are not removable, so it buys only a smaller manifest while putting the hot metadata load path in scope.
4. **Pre-existing limitation, unchanged:** the freshness check compares the pack against the newest input, so an upstream *deletion* leaves a stale entry until a clean build. The per-file layout had the same weakness with orphan files.

Review notes are in the PR comments.


twcclegg:
## Adversarial review (self-review; the multi-agent pass did not complete)

`/code-review max` was run twice. The first attempt was pointed at the wrong diff and its agents then died on a session rate limit; the second read this PR's diff correctly but produced no findings before hitting the same limit. So this is a manual adversarial pass, disclosed as such. Six real issues found and fixed; the commit has been amended.

### 1. Retained-memory regression, 1.9 MB (highest severity)

`ResourcePack` originally materialised every entry into `byte[]` at load. An embedded resource stream is a view over the already-mapped assembly image, so this moved whole data sets onto the GC heap.

Measured with a probe that geocodes three numbers and forces a full collection:

| | retained |
|---|---|
| base (`origin/main`) | 1511 KB |
| PR, materialising | **3430 KB** |
| PR, after fix | 1526 KB |

A pack now holds only its directory and reads entries on demand. The repo's `--retained-memory` audit passes either way (1219 KB against a 1600 KB budget) because it never touches the geocoder — so CI would have shipped this.

### 2. `no` as a property value failed with an unrelated ILLink error

MSBuild condition operators coerce booleans and MSBuild's idea of a boolean includes `yes/no` and `on/off`, so `'no' == 'false'` is **true** — my first validation attempt could never fire for it. ILLink is stricter and takes only the literal strings, so `no` reached it as a malformed feature argument: `IL1030: Invalid argument for '--feature' option`, naming neither the property nor this package.

Values are now normalised to the literal ILLink needs, and a genuine non-boolean is rejected as **PN1002**. Verified: unset → full size; `false`/`False`/`no`/`off` → removal; `banana` → PN1002; `true`+`false` → PN1001.

### 3. I was wrong that the SDK auto-embeds `ILLink.Substitutions.xml`

I claimed the SDK picks it up by convention and that the explicit `<EmbeddedResource>` was redundant. It does not — removing the line leaves the resource absent from the assembly and trimming silently stops working. The earlier CS1508 came from my own duplicate entry. The explicit line stays, and `SubstitutionsFileNamesOnlyResourcesThatExist` was verified to fail when it is removed.

### 4. Exception type narrowed unnecessarily

The time zone mapper previously threw `MissingMetadataException` for an absent resource; the guards replaced it with a bare `InvalidOperationException`, so anyone catching the specific type would have missed it. All three guards now throw `MissingMetadataException`, which derives from `InvalidOperationException` and whose own summary is "resource missing".

### 5. Whole-file diff from line endings

`PhoneNumbers.csproj` is CRLF; my edits rewrote it as LF, turning a ~100-line change into 433. Restored — the diff is now honest.

### 6. Cosmetic

Two double blank lines left by removed helpers; one comment that overstated the atomicity guarantee relative to its siblings; a stale `lib/DumpLocale.java` comment.

### Checked and found sound

Concurrent reads (directory is read-only post-construction, a fresh stream per read); `Lazy` exception caching on all three `GetInstance` paths produces the clean message rather than a `TypeInitializationException`; `dotnet clean` still removes the generated dirs; nothing in `.github/` or `lib/` references the old per-file layout; the pack `.tmp` file cannot be picked up by the embedding globs.

### Known limitation, pre-existing

The freshness check compares the pack against the newest input, so an upstream **deletion** leaves a stale entry until a clean build. The per-file layout had the same weakness with orphan files; not a regression, and CI always builds clean.

twcclegg:
## Multi-agent review: 1 of 10 agents survived, findings applied

The rerun against this PR launched 10 finder angles. Nine died on the session limit (resets 3pm); **"Simplification angle" completed and returned 8 findings**, all of which I verified and all of which held up. Applied in the amended commit — net **−77 lines** across 10 files.

### Applied

1. **`PhoneNumbersFeatures` was mostly ceremony.** The two `AppContext` properties, their `[FeatureSwitchDefinition]` attributes and the two switch-name consts existed solely to pick between two wordings. Nothing read them to decide behaviour. Worse, the `#if NET9_0_OR_GREATER` guard meant net8.0 — a trimmable TFM this package ships — never got the attribute anyway. Collapsed to one message per data set covering both cases: **107 → 57 lines**, and four copies of the resource names are gone.

2. **Dead throw in the time zone mapper.** It scanned `GetManifestResourceNames()` for a name it can compute, which made the follow-up `GetManifestResourceStream(mapFile) ?? throw` unreachable. Now names the resource exactly — removing the last pattern-based resource lookup, which matters because exact names are precisely what `ILLink.Substitutions.xml` couples to.

3. **Guard duplicated at both `PrefixFileReader` construction sites.** Moved into the constructor that owns the pack; `IsDataTrimmed` deleted. The agent's argument is the one that convinced me: an internal property whose only contract is "every caller must remember to check me" means the *next* prefix-backed data set silently gets the empty-string behaviour this PR exists to eliminate.

4. **`BuildAll`'s `Func<int>[]` + rc plumbing guarded a case that cannot occur** — every `Build*` returns 0 on all paths and signals failure by throwing. Straight-line now, with a real stack trace instead of a lambda. It also surfaced three spellings of "make my output directory", one of which (`GetDirectoryName` without `GetFullPath`) throws for a bare relative filename — fine from MSBuild, a crash for a human running the tool. One `EnsureDirectoryFor` helper.

5. **`UnknownKind` still advertised 4 of the 9 kinds.** A typo'd `geocodng` got told geocoding is unsupported. Extracted `PrintUsage()`; both callers use it.

6. **`SupportedCountries()` was a pointless iterator** whose `yield return` deferred the trimmed-data throw to first `MoveNext` — so `LocaleData` would surface it as a `TypeInitializationException`. Now expression-bodied and eager. Unused `using System.Linq` dropped.

7. **`ResourcePack.Entry` struct → value tuple.** Six lines of struct plus a five-line comment apologising for a boxed enumerator, for a two-int pair. `Names` is now the concrete `KeyCollection`, so the boxing is gone too.

8. **Build comments described targets that do not exist.** Including one I authored: *"generation runs once on the outer cross-targeting build (DispatchToInnerBuilds)"* — `DispatchToInnerBuilds` appears nowhere in the file; the target hooks `AssignTargetPaths` only. The original comment was wrong and I carried it forward verbatim. Also removed two orphaned `BuildMetadataBuilder`/`<MSBuild>`-task blocks, corrected the stale target list, documented the trailing-`.` quoting trick that nothing explained, and deleted the unused `MetadataBuilderProject` property from both projects.

### Re-verified after the refactor

Clean all-TFM build, no warnings · 1018 tests green · all flag combinations correct · retained-after-geocoding **1529 KB** vs 1511 KB base · byte-identical assemblies across two packs on all three TFMs · `buildTransitive` present in the nupkg · unknown-kind now prints the full list.

Nothing from the surviving agent was rejected. The nine that died covered correctness, conventions, language pitfalls and altitude — worth rerunning after the limit resets if you want those angles before this leaves draft.

Issue #475, opened 2026-09-19
Extensions: make libphonenumber-csharp.extensions the idiomatic C# layer


twcclegg:
# Extensions: make `libphonenumber-csharp.extensions` the idiomatic C# layer

## Why now, and why not before

`libphonenumber-csharp` is a port. Since the early days, Google's Java library has been the source of truth for both behaviour and API shape, and that has served the project well. Every upstream fix ports cleanly, and anyone reading Google's docs can find the same method under the same name here.

The cost is that some things C# developers reasonably expect were never going to be in the core package. The most requested was a non-throwing parse (#116 in 2020, #160 in 2022). The answer has consistently been that core mirrors Java, and it still does. But #160 ended with a better answer than "no": a separate companion package, `libphonenumber-csharp.extensions`, for C#-shaped conveniences that Java doesn't have (#164). It went through six alpha releases over nearly four months (`8.13.2-alpha1` in December 2022 to `8.13.7-alpha1` in March 2023) before going stable in `8.13.8` alongside a System.Text.Json converter. The hope was that early users would shape it, but the alphas drew no issues or feedback. After that, the API went unchanged for three years, again with no issues filed, until this summer's additions: `TryParse` overloads, a `TypeConverter` and a validation attribute (#405), and AOT JSON support and formatting helpers (#440). Usage has stayed modest, about 0.6 % of core's downloads per release.

Three things have changed since then, and together they make this the right time for a proper pass:

1. **The language caught up.** C# 14 (with .NET 10, November 2025) added *extension members*, which let a separate assembly add **static** members to a type it doesn't own. For the first time, `PhoneNumber.TryParse(...)` can live on the real `PhoneNumbers.PhoneNumber` without the core package changing. Before C# 14 there were only two choices: a helper class with a different name, or changing the ported core type. The companion package took the first route. Its helper class, reasonably, was named after the type it helps.
2. **The tooling caught up.** Over the past months this repo gained package validation against a published baseline (#373), nullable annotations and trim/AOT analysis in core, a systematic analyzer review (#374), and a "blocked on a major version" register (#375). Running the companion package through the same scrutiny made its rough edges measurable:

   | What a consumer writes | What happens today (9.0.39) |
   |---|---|
   | `using PhoneNumbers;` + `using PhoneNumbers.Extensions;` | **CS0104**: `PhoneNumber` is ambiguous between the core type and the Extensions helper class. This applies to every DTO, record and `global using`. |
[...]

PR #476, opened 2026-09-20
feat: framework hooks on PhoneNumber (ToString, IParsable, TypeConverter)

twcclegg:
PR 1 of #475. Adds the three hooks .NET frameworks look for **on the type itself**. None can be supplied from the companion extensions package, which is why they belong in core.

All three live in a new `PhoneNumber.Framework.cs` partial, so an upstream sync never has to reconcile them. The only change to the ported `Phonenumber.cs` is the `partial` keyword.

## What's here

- **`ToString()`** — a verbatim port of Java's `Phonenumber.PhoneNumber.toString()`, which this port never picked up. `"Country Code: 44 National Number: 2070313000"` instead of `"PhoneNumbers.PhoneNumber"`. It reads only fields already on the instance, so it never triggers a metadata load and is safe from a debugger.
- **`IParsable<PhoneNumber>` (net8.0+), implemented explicitly** — ASP.NET Core minimal APIs and MVC now bind a `PhoneNumber` route or query parameter, which was a build error (ASP0020) before; generic `T : IParsable<T>` code works too. Explicit implementation adds **no named member**, so it can't shadow the region-aware `TryParse` the extensions package will provide, and it can't win overload resolution against it.
- **An internal `TypeConverter`**, reached through `[TypeConverter]` on the type — `IOptions<T>` / `IConfiguration` binding of a `PhoneNumber` property now works. It silently bound `null` before.
- `PhoneNumberUtil.Parse` gains an XML `<remarks>` pointing at the extensions package's `TryParse`, so it shows up in IntelliSense.

Both parse paths accept international (`+…`) input only, matching `Parse(s, null)`; region-aware parsing means binding a `string`. Under Native AOT the *configuration-binder source generator* still can't bind `PhoneNumber` (SYSLIB1100) — bind a string there. Both are documented in the XML docs.

## Behaviour change to call out at release

Logging a `PhoneNumber` now emits the number's digits where it used to emit a type name. That makes previously-inert log lines **PII**. It matches Java, and it is the point of the parity fix, but it deserves a CHANGELOG headline when this ships.

## Leading zeros

Java keeps `italian_leading_zero` and `number_of_leading_zeros` as separate fields and prints them separately; this port folds both into `NumberOfLeadingZeros`. Java only sets the count when it isn't 1, so printing the count only when it isn't 1 reproduces Java's output for every number `Parse` can produce. Both cases are tested.

## Test notes

`TestPhoneNumberFramework` covers `ToString` (including culture invariance and an empty number), the `IParsable` implementation through a generic constraint, hostile input, the `TypeConverter` round-trip, configuration binding, and real minimal-API and MVC binding over loopback.

The ASP.NET-backed tests are **net10.0-only**, and the `FrameworkReference` is conditioned to match. The reference lands in `runtimeconfig.json`, so including it on net8.0 aborts the entire net8.0 run on any machine that has the base .NET 8 runtime but not the ASP.NET Core 8 one — which is the case for CI (`setup-dotnet` installs 10.x) and for my machine. The hooks themselves are still covered on net8.0 by the `IParsable` constraint and `TypeConverter` tests. Kestrel on an ephemeral loopback port is used instead of `TestServer`, to avoid adding a package reference.

## Verification

- `dotnet build csharp` — all three TFMs, warnings as errors, trim/AOT analyzers: clean.
- `dotnet test csharp/PhoneNumbers.slnx` — 1030 tests, net8.0 + net10.0: all pass.
- `dotnet pack -c Release csharp/PhoneNumbers` — clean against the 9.0.37 package-validation baseline, confirming this is purely additive.


twcclegg:
## Adversarial review pass

Reviewed the diff assuming it was wrong, with everything checked by running code against both this branch and the published 9.0.39 packages. Four defects found, all fixed in ae27b42b; two pre-existing bugs found that are **not** from this PR; one behaviour change that needs a release note.

### Fixed here

1. **The XML docs were wrong about accepted input.** They said international (`+…`) form only, but `Parse(s, null)` also accepts RFC 3966 URIs, so `TryParse("tel:+1-800-555-0100")` returns `true`. Corrected on both the `IParsable` implementation and the `TypeConverter`, and pinned with a test.
2. **`ToString()` parity is narrower than the commit claimed.** It is faithful for anything `Parse` produces, but a number built through `Builder` can reach states Java prints differently, because the port folds `italian_leading_zero` and `number_of_leading_zeros` into one field. `SetNumberOfLeadingZeros(1)` prints `Leading Zero(s): true` where Java prints `Number of leading zeros: 1`; `SetNumberOfLeadingZeros(2)` prints both lines where Java prints only the count. The remarks now say so, and the folded field's output is pinned for 0/1/2/255.
3. **The culture test asserted nothing.** .NET does not substitute digits by culture, so it passed with or without the invariant formatting it was supposed to guard. It now uses a negative country code under `sv-SE`, whose negative sign is U+2212 rather than U+002D, and skips where culture data is unavailable (globalization-invariant mode).
4. **Untested failure modes**, now covered: the `TypeConverter`'s unsupported-type paths, configuration binding of an *invalid* value (`InvalidOperationException` wrapping `FormatException` — loud, where it silently bound `null` before), and RFC 3966 parsing. Per the `writing-tests` skill, both new entry points also take hostile input in `TestPublicApiRobustness`: the converter must fail only with `FormatException`, and `ToString()` must not throw for anything that parsed.

Also verified, and left alone: `ToString()` omits `RawInput` exactly as Java's does, so two instances `Equals` calls different can print identically. Over 200k random pairs the converse never happened — equal instances always print the same.

### Behaviour change worth a release note (beyond the PII one)

**Newtonsoft.Json now serializes `PhoneNumber` as an E.164 string.** Newtonsoft treats a type with a string-convertible `TypeConverter` as a string contract, so:

| | 9.0.39 | this branch |
|---|---|---|
| `JsonConvert.SerializeObject(number)` | throws `JsonSerializationException` (self-referencing loop via `DefaultInstanceForType`) | `"+442070313000"` |
| `JsonConvert.DeserializeObject<PhoneNumber>("\"+44…\"")` | throws | works, and `Equals` the parsed number |
| with `ReferenceLoopHandling.Ignore` | object-shaped JSON that read back as `0/0` — silent data loss | `"+442070313000"` |

That is a fix rather than a regression (the old output could not round-trip), but anyone who stored the object-shaped JSON now gets an exception on read instead of a silently empty number. It also means the extensions package likely does **not** need the Newtonsoft converter recipe planned in #475.

### Pre-existing bugs, not from this PR (reproduced identically on published 9.0.39)

Builder-only states, unreachable via `Parse`, so low priority — but the new `TypeConverter` can surface them in places like configuration binding and Newtonsoft serialization:

- A **negative country code** makes `PhoneNumberUtil.Format` throw `OverflowException`: `CreateBuilder().SetCountryCode(-5).SetNationalNumber(1)` → `OverflowException` on `Format(..., E164)`.
- `CountryCode = 1`, `NationalNumber = ulong.MaxValue`, `NumberOfLeadingZeros = 3` formats to `"+1000"`, silently dropping the number.

Happy to open an issue for these, or add them to #375 if you would rather they ride the major-version window.

### Verification

Build clean on all three TFMs with warnings as errors; **1047** tests pass on net8.0 + net10.0; `dotnet pack -c Release` clean against the 9.0.37 baseline; `PhoneNumber.Framework.cs` at **100% line coverage** for both new types (82/82 and 30/30), against the repo's 90% patch gate.

twcclegg:
## Second review pass: mutation testing, line endings, and re-checking my own claims

Prompted by a peer session's advice after its own review: verify that each guard test actually *fails* when you break what it guards, audit line endings, and check hardest the claims already asserted as verified. That found one hollow guard (fixed in d726df1f) and one unproven doc claim (now proven).

### Mutation testing — do the guards actually guard?

Each behaviour was broken one at a time, the relevant tests run, then reverted:

| # | Mutation | Result |
|---|---|---|
| M1 | `ToString` formats with `CurrentCulture` instead of `InvariantCulture` | **caught** — `ToStringIsCultureInvariant` fails |
| M2 | Drop `[TypeConverter]` from `PhoneNumber` | **caught** — 5 tests fail (converter, `TypeDescriptor`, config binding) |
| M3 | Always print the leading-zero count (drop the `!= 1` rule) | **caught** — 2 parity tests fail |
| M4 | `ToString` throws for a number carrying an extension | **initially only half-caught — see below** |
| M5 | `IParsable.TryParse` returns `true` with a null result | **caught at compile time** (CS8762, the nullability contract) |
| M6 | `TypeConverter` returns null instead of throwing on bad input | **caught** — 2 tests fail |

**M4 exposed a hollow guard.** `ToStringNeverThrowsForAnythingThatParsed` passed while `ToString()` threw on every number with an extension: none of the hostile inputs parsed into a number that had one, so the sweep only ever reached the two unconditional fields. Only the unit test caught it.

Fixed by adding two inputs that parse *with* an extension, and asserting the sweep actually observed an extension and a country code source — so the test fails rather than quietly narrowing if those inputs stop parsing. Verified in both directions: the mutant now fails it, and deleting the two inputs fails it too.

### Claims I had already asserted, re-checked

- **"Adds no named public member."** Confirmed by reflection over the built assembly: `PhoneNumber`'s declared public members contain no `Parse`/`TryParse`, both explicit implementations are private (`System.IParsable<PhoneNumbers.PhoneNumber>.TryParse`, `IsPublic=False`), and `PhoneNumberTypeConverter` is absent from `GetExportedTypes()`.
- **"`ToString()` never loads metadata, so it is safe from a debugger."** This is asserted in shipped XML docs and my first attempt to verify it was **hollow**: I instrumented `MetadataManager.SetMetadataLoader`, which reported 0 loads for `ToString` — but also 0 for a real `Parse`, because that API only wires the alternate-formats and short-number sources, never the main parse path. (Independent support for internalizing it in #375.) Re-verified against the thing that does gate metadata, the lazily created `PhoneNumberUtil` singleton, with a working control:
  - after `ToString()` exercising *every* branch: singleton still `null`, 5,632 bytes allocated;
  - control, after `GetInstance()` + `Parse`: singleton created, 179,616 bytes allocated.
- **Coverage.** Re-measured after these changes: still 82/82 and 30/30 lines, no misses.
- **Test count.** 1047 across net8.0 + net10.0.

### Line endings

No `.gitattributes` here and I made scripted edits to two CRLF files, so this was a live risk. Audited with `file`: `Phonenumber.cs` and `PhoneNumberUtil.cs` are still CRLF-with-BOM as on `main`, and their diffs are 3 and 5 lines rather than whole-file rewrites. The new `PhoneNumber.Framework.cs` is LF, which matches the convention for newer files in that directory (28 of 43 are LF, including the most recently added).

### Unrelated, for your awareness

Two untracked files appeared in the repo root on 20 Sept — `install-debuntu.sh` and `install-debuntu.sh.sha256sum`. They are not mine and are not in any commit on this branch; I left them alone. Worth a look if you don't recognise them.

twcclegg:
## Adversarial review

Nine findings, ranked. Five of them trace to one line — `[TypeConverter(...)]` at `PhoneNumber.Framework.cs:28`. That attribute isn't additive in the way the package-validation baseline reports: it retroactively changes how three separate ecosystems treat a type they already handled (MVC/minimal-API binding-source inference, Newtonsoft.Json contract resolution, `ConfigurationBinder`'s empty-value path). None of it produces a compile error for a consumer, and none of it is reachable from this repo's test suite — which is why the clean `dotnet pack` against the 9.0.37 baseline doesn't speak to it. `ToString()` and `IParsable` are genuinely low-risk by comparison.

---

### 1. `[TypeConverter]` flips MVC / minimal-API binding source from body to query — `PhoneNumber.Framework.cs:28`

`ModelMetadata.IsComplexType` is literally `!TypeDescriptor.GetConverter(ModelType).CanConvertFrom(typeof(string))`, so a string-capable converter makes `PhoneNumber` a *simple* type.

Concrete break: a consumer on the current release who registered `PhoneNumbers.Extensions.PhoneNumberJsonOptions` / `PhoneNumberConverter` and wrote

```csharp
[ApiController] public IActionResult Post(PhoneNumber n)
```

had `n` inferred as `[FromBody]` and deserialized from JSON. After upgrading, `InferParameterBindingInfoConvention` sees a non-complex type and infers `BindingSource.Query`, so the JSON body is ignored and `n` arrives null / the request 400s. The same flip happens in minimal APIs — `ParameterBindingMethodCache` now finds a `TryParse`, so `app.MapPost("/x", (PhoneNumber n) => …)` stops binding from the body.

Nothing in the diff, the tests, or the PR body covers this. At minimum it needs a release-note callout alongside the PII note.

### 2. `[TypeConverter]` changes Newtonsoft.Json's shape for `PhoneNumber` — `PhoneNumber.Framework.cs:28`

`DefaultContractResolver.CreateContract` calls `CanConvertToString(t)` — true for any non-default converter whose `CanConvertTo(typeof(string))` is true — *before* falling through to `CreateObjectContract`, and returns `CreateStringContract` instead.

So `JsonConvert.SerializeObject(number)` goes from `{"CountryCode":44,"NationalNumber":2070313000,…}` to `"+442070313000"`, on both read and write. Deserializing an existing stored payload into a `PhoneNumber` then fails or yields a different value. Combined with finding 4, a Newtonsoft round-trip of a number with an extension silently loses it.

System.Text.Json ignores `TypeConverter`, so the existing `PhoneNumberConverter` / `PhoneNumberJsonContext` tests can't surface this.

### 3. Empty configuration value now throws instead of binding null — `PhoneNumber.Framework.cs:150`

`ConfigurationBinder.TryConvertValue` short-circuits `""` only for `Nullable<T>`. `PhoneNumber` is a reference type, so `"Sms:From": ""` in appsettings.json goes straight to `converter.ConvertFromInvariantString("")` → `Parse("", null)` → `NumberParseException` → `FormatException` → `InvalidOperationException` out of `.Get<SmsOptions>()`. A blank optional setting becomes a startup crash; before this PR it left the property null.

`ConfigurationSurfacesAnInvalidValueAsAnError` only pins `"junk"`, where throwing is right. Suggest `if (stringValue.Length == 0) return null;` before the `Parse` call, with a test.

### 4. The converter's round-trip is lossy — `PhoneNumber.Framework.cs:170`

`PhoneNumberUtil.cs:1209-1214` takes an early exit for E164 with the comment *"Extensions are not formatted"*. `ConvertFrom` happily accepts input carrying an extension — the new hostile input `"tel:+18005550100;ext=99"` (`TestPublicApiRobustness.cs:57`) parses into a number with `Extension` set — but `ConvertTo` emits `"+18005550100"`. `PreferredDomesticCarrierCode`, `CountryCodeSource` and `RawInput` go too.

The summary at line 133 claims the converter lets infrastructure "round-trip one". `TypeConverterRoundTripsThroughE164` only exercises a number with none of those fields set. Either use `RFC3966` for `ConvertTo`, or say plainly in the doc what is lost.

### 5. Duplicates the existing public converter and orphans its doc — `PhoneNumber.Framework.cs:145`

`csharp/PhoneNumbers.Extensions/PhoneNumberTypeConverter.cs` is the same name and a near-identical body. Its summary still says *"Not applied automatically — register it where needed, e.g. `TypeDescriptor.AddAttributes(...)`"*, which is now false: one **is** applied automatically, and a consumer following that doc replaces the built-in with a near-clone.

The two also already disagree on the failure message (`"is not a valid phone number."` vs `"is not a valid international phone number."`), so which text a consumer sees depends on whether they ever called `AddAttributes`. Either delete/deprecate the extensions copy and fix its doc, or have core delegate so the behaviour lives in one place.

Findings 1, 2 and 5 together are the argument that the converter specifically belongs in `PhoneNumbers.Extensions`, where an opt-in registration carries none of the retroactive risk. (`AGENTS.md:81` also asks for per-change sign-off on public additions to core — presumably settled here, but worth stating in the PR body.)

### 6. `IParsable<PhoneNumber>.Parse` throws the wrong exception type — `PhoneNumber.Framework.cs:104`

`IParsable<TSelf>.Parse` documents `FormatException` for unparseable input and `ArgumentNullException` for a null `s`. This implementation lets `NumberParseException` — derived straight from `Exception` — escape for both; `ParseHelper` maps null to `NOT_A_NUMBER` at `PhoneNumberUtil.cs:2865`. So generic code like

```csharp
static T? Or<T>(string s) where T : IParsable<T>
{
    try { return T.Parse(s, null); } catch (FormatException) { return default; }
}
```

does not catch. The `TypeConverter` ten lines below *does* honour the contract, so the two hooks in the same file behave differently. `IParsableThrowsOnInvalidInput` pins the divergence rather than flagging it.

Trade-off worth weighing: wrapping in `FormatException` hides `NumberParseException.ErrorType` from callers who want it. Wrapping with the original as `InnerException` (as the converter does) gets both.

### 7. The `ToString()` PII change isn't recorded anywhere outside this PR body — `PhoneNumber.Framework.cs:63`

`ILogger`/Serilog `{Number}` placeholders, `$"{number}"` and exception messages previously emitted `"PhoneNumbers.PhoneNumber"`; they now emit the subscriber's digits, so previously-inert log lines become PII at rest. The PR body acknowledges it, but CHANGELOG entries are written at release time and this commit subject won't convey it to anyone reading the release notes.

Second, unmentioned instance of the same class: `PhoneNumberTypeConverter.ConvertFrom` interpolates the raw input into the `FormatException` message, which is exactly what `ConfigurationBinder` wraps and logs on a binding failure.

### 8. `ToStringIsCultureInvariant` silently no-ops — `TestPhoneNumberFramework.cs:105`

Two `return` paths (no `sv-SE` culture data, or a runtime whose `NegativeSign` isn't U+2212) make the test pass without reaching a single assert, and it's reported green. A regression that leaked `CurrentCulture` into `ToString` would go unnoticed on that runner. `Assert.Skip` / `Skip.If` makes the drop visible.

### 9. The hostile-input sweep guards two of `ToString`'s four optional branches — `TestPublicApiRobustness.cs:240`

`sawExtension` and `sawCountryCodeSource` are asserted, but nothing forces `HasNumberOfLeadingZeros` or `HasPreferredDomesticCarrierCode`. A `ToString` mutated to throw in either branch still passes — which is precisely the failure mode the test's own comment says it was written to close for the extension branch. An IT national number as an input plus a `sawLeadingZeros` flag would close half of it.

---

### Checked and found clean

- **Java parity of `ToString()`** — field order, `CountryCodeSource` naming, `RawInput` exclusion, and the `italian_leading_zero` / `number_of_leading_zeros` folding. `HasNumberOfLeadingZeros => _numberOfLeadingZeros != 0` (`Phonenumber.cs:45`), and upstream sets the count only when it isn't 1, so the `!= 1` guard does reproduce Java for every number `Parse` can produce. `Parse` sets the source only under `keepRawInput` (`PhoneNumberUtil.cs:2458-2460`, `2511-2512`), so `ToStringMatchesJavaFormat` expecting no source line is right.
- **`TryParse` catching only `NumberParseException` is safe** — `MAX_INPUT_STRING_LENGTH` (250) and the `MAX_LENGTH_FOR_NSN` guard both precede the `ulong.Parse`, so no `OverflowException` can escape.
- **No in-repo caller relies on the old `object.ToString()`** — every `*.ToString()` hit in the library is a `StringBuilder`, and nothing in the Blazor demo renders a `PhoneNumber` directly.
- **`internal sealed` converter is reachable** via `Activator.CreateInstance` (public implicit ctor; type visibility isn't checked), and `CultureInfo.InvariantCulture` on the `Append` calls is necessary rather than defensive — `StringBuilder.Append(int)` is culture-sensitive on netstandard2.0.
- **Explicit interface implementation genuinely adds no named member**, and TFM parity holds (the net8.0 surface is a superset of netstandard2.0).
- **The net10.0-only `FrameworkReference`** matches the `#if NET10_0_OR_GREATER` guards exactly, and `app.Urls` is correctly populated by Kestrel's `AddressBinder` after an explicit `Listen(IPAddress.Loopback, 0)`.
- **Adding two parseable strings to `HostileInputs`** doesn't disturb the other tests in that class — `AssertOverInputs` tolerates success.
- **No CHANGELOG entry is expected in-PR** — entries are written at release time by `lib/update-changelog.sh`.

One doc item: `docs/api-differences-from-java.md` doesn't mention `ToString`, and this PR closes that divergence.

<sub>🤖 Review assisted by [Claude Code](https://claude.com/claude-code)</sub>

